PR #195, opened 2014-11-09
Make Home and End keys work in overview

nchachereau:
Sometimes, in the overview window, one wants to go rapidly to the first or to the last entry (especially since the latter is often the current activity).
This small patch fixes this.

(I also fixed some typos in comments along the way, please tell me if this is okay or not!)


toupeira:
Thanks!

PR #212, opened 2014-12-31
Added gir1.2-gconf-2.0 as dependency (Debian)

mbojan:
As in here http://choorucode.com/2014/10/06/cannot-import-name-gconf-error


toupeira:
Thanks!

PR #218, opened 2015-02-05
Fixed export unicode error

OlivierBerro:
Fixed encoding error when eporting data.  

_Error_

>  File "/usr/local/lib/python2.7/dist-packages/hamster/reports.py", line 162, in **init**
>    self.csv_writer.writerow([h for h in headers])
> UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 7: ordinal not in range(128)

### DIFF
--- a/src/hamster/reports.py
+++ b/src/hamster/reports.py
@@ -44,6 +44,10 @@ from calendar import timegm
 
 from StringIO import StringIO
 
+# fix encoding on export
+reload(sys)
+sys.setdefaultencoding("utf-8")
+
 def simple(facts, start_date, end_date, format, path = None):
     facts = copy.deepcopy(facts) # dont want to do anything bad to the input
     report_path = stuff.locale_from_utf8(path)


nchachereau:
This is a rather ugly way to handle this. IMHO, we should sort out our unicode / string problems, rather than reloading sys and setting the default to UTF-8, which could lead to other problems. What do others think?

FelixSchwarz:
I completely agree with @nchachereau  (without any special authority for this project of course). If things break like that it seems to be a symptom of ill-defined encoding and this should/needs to be resolved for Python 3 anyway.

axmachado:
Hello, people...

I think I've found a better way to do it. Just submitted an pull request with my solution. Please take a look at it.

elbenfreund:

ederag:
Thanks a lot.
This utf-8 issue has hopefully been fixed by the [python3 migration](https://github.com/projecthamster/hamster/releases/tag/v2.1.1).

Issue #213, opened 2015-01-21
Alt-Shortcut Letter Duplicated on 'Tracking' menu and 'Start tracking' button


Orbixx:
As per the subject, to reproduce, just open Hamster and hold 'Alt' and you can see the underscored 'T' in "Tracking" and "Start tracking". It defaults to "Tracking" if you hit 'T'.

Orbixx:
Pull request here: https://github.com/projecthamster/hamster/pull/225

PR #224, opened 2015-03-15
Fixes mnemonic accelerator key conflict of "Tracking" and "Start Tracking" in today.ui

Orbixx:
Switched the underscore shortcut character of "Start Tracking" from "T" to "S", this no longer conflicts with "Tracking" and allows quick keyboard-only usage of Hamster.

This resolves issue #213


Orbixx:
Translations weren't *quite* updated, will submit another pull request.

PR #225, opened 2015-03-15
Fixes mnemonic accelerator key conflict of "Tracking" and "Start Tracking" in today.ui

Orbixx:
Switched the underscore shortcut character of "Start Tracking" from "T" to "S", this no longer conflicts with "Tracking" and allows quick keyboard-only usage of Hamster.

Please check some of the translation files as I think my editor may have re-encoded them, causing them to completely replace one another with git.

Have also fixed a translation error in nl.po

This resolves issue #213

### DIFF
--- a/po/ro.po
+++ b/po/ro.po
@@ -392,8 +392,8 @@ msgid "S_witch"
 msgstr "Com_utare"
 
 #: ../data/today.ui.h:11
-msgid "Start _Tracking"
-msgstr "Porneș_te pontarea"
+msgid "_Start Tracking"
+msgstr "Porne_ște pontarea"
 
 #: ../data/today.ui.h:12
 msgid "Start new activity"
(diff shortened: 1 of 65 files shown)


po/ro.po
```diff
@@ -392,8 +392,8 @@ msgid "S_witch"
 msgstr "Com_utare"
 
 #: ../data/today.ui.h:11
-msgid "Start _Tracking"
-msgstr "Porneș_te pontarea"
+msgid "_Start Tracking"
+msgstr "Porne_ște pontarea"
```
toupeira: Not sure if this will work with a normal `s`

toupeira:
Hey there, sorry for the slow response!

Your editor seems to have converted the CRLF linebreaks which is appreciated :)
You can append `?/w=1` to a GitHub diff to ignore whitespace changes, see https://github.com/projecthamster/hamster/pull/225/files?w=1#diff-8e8e58140f1c5c0a5e5484b9dd70e339

PR #226, opened 2015-03-27
In a python3 or mixed environment specifiy the major version to use

fosero:
Currently hamster doesn't work in a python3 environment because of these 2 babies.


toupeira:
Thanks!

PR #241, opened 2015-06-25
Incorrect tags in html reports

istarii:
When generating HTML reports , hamster incorrectly handled tags.
Such was before the change
before

so after a change
after


toupeira:
Looks good, thanks a lot!

PR #259, opened 2015-10-08
Associate tasks to specific workspaces and follow changes automatically

aurelien-naldi:
I added a simple mechanism to save/restore previous tasks and used it to:
* resume tasks which have been interrupted by screen lock
* change the current task upon workspace changes

It still lacks proper setting (and their GUI) but it "works for me" :)

### DIFF
--- a/src/hamster/lib/desktop.py
+++ b/src/hamster/lib/desktop.py
@@ -24,6 +24,7 @@ from gi.repository import GObject as gobject
 
 
 from hamster import idle
+from hamster import workspace
 from hamster.lib.configuration import conf
 from hamster.lib import trophies
 import dbus
@@ -47,6 +48,10 @@ class DesktopIntegrations(object):
         self.idle_listener.connect('idle-changed', self.on_idle_changed)
 
         gobject.timeout_add_seconds(60, self.check_hamster)
+
+        self.use_workspace_changes = True # TODO: should be a proper setting
+        self.workspace_listener = workspace.WnckWatcher()
+        self.workspace_listener.connect('workspace-changed', self.on_workspace_changed)
 
 
     def check_hamster(self):
@@ -124,6 +129,15 @@ class DesktopIntegrations(object):
             self.restore('__paused__', delete=True)
 
 
+    def on_workspace_changed(self, event, previous_workspace, current_workspace):
+        if not self.use_workspace_changes:
+            return
+        
+        if previous_workspace is not None:
+            self.save_current( previous_workspace )
+        self.restore(current_workspace)
+
+
     def on_conf_changed(self, event, key, value):
         if hasattr(self, "conf_%s" % key):
             setattr(self, "conf_%s" % key, value)
@@ -141,7 +155,7 @@ class DesktopIntegrations(object):
 
     def save_current(self, key):
         current = self.get_current()
-        if current:
+        if current is not None:
             self._saved[key] = current
         elif key in self._saved:
             del self._saved[key]
--- a/src/hamster/lib/desktop.py
+++ b/src/hamster/lib/desktop.py
@@ -33,6 +33,7 @@ class DesktopIntegrations(object):
     def __init__(self, storage):
         self.storage = storage # can't use client as then we get in a dbus loop
         self._last_notification = None
+        self._saved = {}
 
         dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
         self.bus = dbus.SessionBus()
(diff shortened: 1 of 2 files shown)


src/hamster/workspace.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    __gsignals__ = {
+        "workspace-changed": (gobject.SIGNAL_RUN_LAST, gobject.TYPE_NONE, (gobject.TYPE_PYOBJECT,gobject.TYPE_PYOBJECT))
+    }
+    def __init__(self):
+        gobject.GObject.__init__(self)
+
+        try:
+            self.bus = dbus.SessionBus()
+        except:
```
toupeira: Please only catch the specific exceptions you expect.

src/hamster/workspace.py
```diff
@@ -0,0 +1,51 @@
...
+            return 0
+        wnck.Screen.get_default().connect("active-workspace-changed", self.on_workspace_changed)
+
+
+    def on_workspace_changed(self, screen, previous_workspace):
+        try:
+            workspace_id = screen.get_active_workspace().get_number()
+            if previous_workspace:
+                previous_workspace = previous_workspace.get_number()
+        except:
```
toupeira: Same here.

src/hamster/lib/desktop.py
```diff
@@ -108,13 +114,66 @@ def notify_user(self, summary="", details=""):
...
+
+
+    def get_current(self):
+        facts = self.storage.get_todays_facts()
+        if facts:
+            last = facts[-1]
+            if not last['end_time']:
+                name = last['name']
+                category = last['category']
+                return name,category
```
toupeira: Hmm this should also include tags and other metadata, basically the whole `Fact` instance.

toupeira:
Thanks for your contribution! To merge this I would definitely like:

- Proper GConf settings, the old `workspace_tracking` preference supported the values `memory` (remember activity by workspace, i.e. your behaviour) and `name` (switch to specific activites as defined in the old `workspace_mapping` preference). I propose we keep this as a string list but ignore values other than `memory` for now.
- Disable the new settings by default so the behaviour doesn't change, which also means we don't need to update the preferences UI just yet (though that would be welcome as well of course)
- The save/restore mechanism looks like a good source for obscure bugs, so we should probably make this a bit more reliable and find edge-cases to test

Let me know if you need any pointers.

src/hamster/workspace.py
aurelien-naldi: As I'm not familiar with dbus, I copied this from idle.py, but all other places in hamster where a session bus is retrieved are unprotected, so I'm taking out the try/except here.

src/hamster/workspace.py
aurelien-naldi: Here I'm not so sure about what could go wrong, I guess the main source of actual problem is at import time: if wnck is missing. I will also remove this try/catch
On a related note, I wonder what will happen to wnck in a wayland world.

aurelien-naldi:
Thanks for your feedback!

I agree on the settings part but I first wanted to know if it had a chance of getting in. I also want to add a "auto resume" setting to the lock screen part.

For the save/restore part, I wanted to start by adding a simple API for it, but indeed the implementation needs to be solid, maybe some additions to the storage object could help: is there a way to reuse the existing fact object to start a new slot? The gnome-shell extension has buttons to restart one of the previous tasks, this code could be shared.

elbenfreund:

GeraldJansen:
Considering #470, #493 and #494, this PR should probably be closed.

mwilck:
> Considering #470, #493 and #494, this PR should probably be closed.

Yes, in view of #494, this isn't going to be implemented any time soon. We'd need to get the desktop integration back first.

Do we have an option to save this as possible feature / future work?

GeraldJansen:
> Do we have an option to save this as possible feature / future work?

Aside from leaving it open, another option would be to open a new feature request issue in whatever project/fork undertakes reviving desktop integration and linking back here. AFAIK closed requests don't disappear, they are just less visible.

mwilck:
> AFAIK closed requests don't disappear, they are just less visible.

You might as well say "invisible". Who scans closed issues? We should differentiate between "will never implement" and "won't implement just now".

ederag:

matthijskooijman:
W00ps, i was a little overzealous in cleanup @ederag's comment. Turns out it still relevant, since there is a "Pending check" for CLA that can probably never realized. Here's the deleted comment:

> There will never be any CLA on this code. Please disregard the wrong checks (#589).

Sorry for the noise.

Issue #221, opened 2015-02-13
ImportError: No module named hamster


auipga:
Using Arch Linux
```
$ uname -a
Linux auipga 3.18.6-1-ARCH #1 SMP PREEMPT Sat Feb 7 08:44:05 CET 2015 x86_64 GNU/Linux
```
Installed dependencies:
```
python-dbus python-xdg gnome-control-center python2-libgnome python2-gconf python2-gnomecanvas python2-wnck intltool gnome-doc-utils docbook2x
python2-pysqlite (from AUR)
```
```
$ ./waf configure build --prefix=/usr/local
Checking for program python              : /usr/sbin/python2 
Checking for Python version >= 2.4.2     : ok 2.7.9 
Checking for program glib-genmarshal     : /usr/sbin/glib-genmarshal 
Checking for program glib-mkenums        : /usr/sbin/glib-mkenums 
Checking for program dbus-binding-tool   : /usr/sbin/dbus-binding-tool 
Checking for program docbook2man         : /usr/sbin/docbook2man 
Checking for program xml2po              : /usr/sbin/xml2po 
Checking for program xsltproc            : /usr/sbin/xsltproc 
Checking for program msgfmt              : /usr/sbin/msgfmt 
Checking for program intltool-merge      : /usr/sbin/intltool-merge 
Checking for program xml2po              : /usr/sbin/xml2po 
'configure' finished successfully (0.041s)
Waf: Entering directory `/home/auipga/hamster/build'
( files 1 to 200, everything looks ok )
Generating and caching the translation database
Merging translations into default/data/hamster-time-tracker.schemas.
Waf: Leaving directory `/home/auipga/hamster/build'
'build' finished successfully (3.405s)
```
```
$ sudo ./waf install
(ends with)
'install' finished successfully (0.154s)
```

Till now everithing looks ok.
Running hamster fails.
```
$ hamster
Traceback (most recent call last):
  File "/usr/local/bin/hamster", line 30, in <module>
    from hamster import client, reports
ImportError: No module named hamster
```

toupeira:
Please try again with `sudo ./waf install`.

auipga:
I did. (changed the first post)

billyworm:
I think have the (quite) same issue with Ubuntu 14.04:

``` $ uname -a
[...]

PR #276, opened 2016-01-04
Updated README's build instructions to avoid lib import error

Orbixx:
As shown in #221 targeting prefix /usr/local appears to break library imports on various distros. I don't think any other systems that have already been building to /usr/local would break if they changed to /usr

Distro-specific packages can of course build to whatever dir works for them anyway, too.


FelixSchwarz:
sounds sensible

Issue #12, opened 2012-09-24
make it easier to enter tags


tstriker:
right now you can't enter nor see the tags through the extension at all, so that's for one, but there is a demand for automated or at least auto-suggested sets of categories. not so sure about that path, but the requirement is a valid one so something could be done.

originals: 
https://bugzilla.gnome.org/show_bug.cgi?id=641243
https://bugzilla.gnome.org/show_bug.cgi?id=666782

PR #303, opened 2016-03-02
Add independent completion for categories and tags

CendioOssman:
They are often not bound to specific activities so make sure all
recent ones are suggested. This is even more important when entering
a previously unseen activity.

### DIFF
--- a/src/hamster/widgets/activityentry.py
+++ b/src/hamster/widgets/activityentry.py
@@ -44,6 +44,12 @@ def extract_search(text):
         search += "@%s" % fact.category
     if fact.tags:
         search += " #%s" % (" #".join(fact.tags))
+    # Add back any start of a category or tag as the Fact parser will
+    # strip off anything incomplete
+    if text.endswith("@"):
+        search += "@"
+    elif text.endswith("#"):
+        search += "#"
     return search
 
 class DataRow(object):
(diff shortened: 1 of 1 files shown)


CendioOssman:
I assume the same fix is needed for the extension.

axmachado:
It looks interesting... will try to test it.

elbenfreund:

ederag:
After a tentative merge:
https://github.com/ederag/hamster/tree/CendioOssman-completion
it increases the code duplication (with `hamster.lib.Fact` and `parse_fact`)
already present in the common ancestor.
And it might be because of wrong choices during merge
(please help if it is the case),
but it is impossible to search specifically for a tag by typing `#` as the first letter,
while this is possible in current master.
It does not find tags in the auto-completion configuration,
but both versions find "unseen" activities.

So it seems that the issues this PR was intending to solve
have been resolved in another way by PR #363.
Please comment here if I missed anything.

Your work was useful though,
because the separation between activity, category and tags searches is very interesting,
and besides, it has surely been silently been used by others meanwhile.

ederag:
related: #12

CendioOssman:
I think this is still needed in some form. This is just based on looking at the current code though, I haven't tested anything.

Please re-open the PR and I'll update the branch to something that works with the existing code.

ederag:
Since another PR had to be used to complement this one,
it would be better to keep this one closed, with its current history.
Please open a new PR, linking to this one.

Some comments before you start coding:
1) Please start by identifying clear use cases where the current code fails.
    We need to gather a bunch of tests anyway.
2) Searching for category or tags separately *is* possible now:
  just start the search with `@` or `#` respectively
3) What does not work well is to search for two tags,
    because they have to be given in the correct order to be found.
    The search for tags should not depend on the order.
    Or could it be useful for some users to have some "priority" order ?
4) There is still a lot of code duplication with `hamster.lib.__init__.py`,
     that should be removed in the process.
    (search should be based on facts fields)

Issue #286, opened 2016-01-21
Activities with accent marks in name 


miguelcordero:
If I create a activity with acute mark or accent mark in their name, like "Exposición", hamster doesn't permit save or modify their description. 
The Save button doesn't work when i type something in description box and window doesn't close. 
If I type in Title box, after the @,  button Save works and closes the window but not saves the new content. 

I run Debian 8.2, and i test with hamster_2.0-rc1-2_all.deb and building from source with master release also.

ederag:
Thanks a lot.
This utf-8 issue has hopefully been fixed by the [python3 migration](https://github.com/projecthamster/hamster/releases/tag/v2.1.1).

Issue #239, opened 2015-05-21
Report fails when localized date contains non-ASCII characters (e.g. Février)


jgribonvald:
I'm having problems to generate a report for "Février" (February) month, and I'm having the same problem for any day on this month. Do you think that can be due to french month name with accents ?

Any help to resolve for this bug would be appreciated.

Thanks

nchachereau:
> Do you think that can be due to french month name with accents ?

Unfortunately, yes, it does. (`hamster export 2015-02-01 2015-02-28` will show you the error). Unicode was difficult to handle in Python2, and Hamster does not do it very well. (See also #153).

I have started looking into this. In the meantime, if this is urgent, you could try doing the same as pull request #218 : adding the following lines near the start of reports.py

```py
reload(sys)
sys.setdefaultencoding("utf-8")
```

jgribonvald:
Thanks, #218 is solving the problem !

ederag:
Thanks a lot.
This utf-8 issue has hopefully been fixed by the [python3 migration](https://github.com/projecthamster/hamster/releases/tag/v2.1.1).

Issue #153, opened 2014-04-01
Encoding errors in non-ASCII locales


gflohr:
Current git checkout with python 2.7.5 on gentoo, running a Gnome session in the locale de_DE.UTF-8:

  $ /usr/bin/hamster list

  Traceback (most recent call last):
    File "/usr/bin/hamster", line 391, in <module>
      getattr(hamster_client, command)(_args)
    File "/usr/bin/hamster", line 254, in list
      self._list(start_time, end_time)
    File "/usr/bin/hamster", line 313, in _list
      print fact_line.format(_*headers)
  UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 1: ordinal not in range(128)
  $

The string "Activity" from in /usr/bin/hamster line 313 translates to "Tätigkeit" in German, and 0xe4 is the code point for the "ä".  The issue vanishes by calling gettext.install() in src/hamster/lib/i18n.py with the "unicode" argument set to False. 

Now without the "list" command:

  $ /usr/bin/hamster
   ...
  File "/usr/lib64/python2.7/site-packages/hamster/lib/stuff.py", line 91, in format_range
    title = (u"%(start_B)s %(start_d)s – %(end_B)s %(end_d)s, %(end_Y)s") % dates_dict
  UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)

After changing the u"" strings into byte strings without prefix, I get this:

```
...
self.label.set_markup('<b>%s</b>' % stuff.format_range(start_date, end_date).encode("utf-8"))
```

  UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)

Note that 0xc3 is the code point "Ã".  So this is obviously a double encoding error for utf-8.  The problem vanishes after removing the encode('utf-8') from set_range() in src/hamster/widgets/dates.py:

```
def set_range(self, start_date, end_date=None):
    end_date = end_date or start_date
    self.start_date, self.end_date = start_date, end_date
    #self.label.set_markup('<b>%s</b>' % stuff.format_range(start_date, end_date).encode("utf-8"))
    self.label.set_markup('<b>%s</b>' % stuff.format_range(start_date, end_date))
```

After those modifications the window shows up and the application seems to work.

I can send you a real patch if you want but to me it looks like a crude workaround for a more general problem.

stedybg:
[...]

PR #308, opened 2016-03-13
Solve issues relates do character encoding

axmachado:
Solve issues #286, #239 and #153.

The approach here is to use unicode string all the way, and encode it to the system default encoding only when writing to console or file. It changes a bit how things are done in "stuff.py" and "reports.py", formatting all strings in unicode before encoding to the system charset.

I removed some hardcoded "utf-8" encodings, converting it to Unicode strings.

For me, it solved all the encoding issues that where making hamster unusable in my system. I hope it work for you all too.

There is some risk of this patch breaking something related to bug 562298 (encoding issue with Japanese locale). I think it will not break, but I was unable to test it. If someone that uses Japanese locale can test, I will be glad.

I can spare some more hours to work with hamster, so, if anyone thinks any bug is really annoying, please, let me know.

### DIFF
--- a/src/hamster-cli
+++ b/src/hamster-cli
@@ -27,6 +27,8 @@ import optparse
 import re
 import datetime as dt
 
+from locale import getdefaultlocale, getpreferredencoding
+
 from hamster import client, reports
 from hamster.lib import Fact, stuff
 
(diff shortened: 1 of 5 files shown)


dClauzel:
When generating a web report, the names with accents are broken.

Example, for the month of december "déc." becomes "dÃ©c.".

ederag:
Thanks a lot.
This utf-8 issue has hopefully been fixed by the [python3 migration](https://github.com/projecthamster/hamster/releases/tag/v2.1.1).

PR #342, opened 2017-03-03
hamster: use gi.require_version for every gi module

mwilck:
This should silence all
"PyGIWarning: XXX was imported without specifying a version first."
log messages.

Users have complained about these warnings repeatedly. This patch bundles all gi version requirements in one place, lib/gi_versions.py, and just imports that were necessary.

FTR: I understand that you're not putting maintenance efforts in this code base any more. However I wanted to share this with you anyway. I've taken ownership of hamster-time-tracker for openSUSE. In this role, I need and want to share all fixes I make for the upstream package with upstream. I'll checkout the new code base at some later time.

### DIFF
--- a/src/hamster-cli
+++ b/src/hamster-cli
@@ -143,6 +143,7 @@ class HamsterClient(object):
             getattr(server, window_name)()
         else:
             print "Running in devel mode"
+            from hamster.lib import gi_versions
             from gi.repository import Gtk as gtk
             from hamster.lib.configuration import dialogs
             getattr(dialogs, window_name).show()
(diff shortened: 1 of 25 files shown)


elbenfreund:
Hi Martin, great to hear from you.
One of the reasons we have not simply shut down this repository/tracker is so people can still share their thoughts and solutions on the old codebase for the time being. So thank you for that :+1: 

This issue this PR addresses is solved in the new codebase and we should certainly stay in touch to work together in bringing the 'new hamster' to openSUSE once we are ready to do so.

ederag:
As @elbenfreund said, this seems to have been fixed before 2.1.1.
For interested openSUSE users, @mwilck is [reviewing](https://build.opensuse.org/request/show/653434) an updated version of the package.
Closing, but if I missed anything, please ask here for reopening.

PR #347, opened 2017-06-15
Fix a divide-by-zero error when rendering bars with zero width

bcdickinson:
In the stats view, if there are any zero-width bars to display, there's a ZeroDivisionError, which causes some bizarre rendering glitches. I noticed this when I had an unfinished task at the end of a day against a tag.


ederag:
Thanks for contributing !
This is fixed in v2.1.1.

PR #302, opened 2016-03-02
Some minor layout fixes

CendioOssman:

PR #348, opened 2017-08-07
Hours sum and percentage of viewed time period in Totals view

HWiese1980:
I needed to see the overall hours and the percentage of activities, categories and tags in the Totals view as hard numbers. I've added the values to the label of the corresponding bar graph in the Totals view. This might be of interest for others as well, so feel free to merge this contribution if you like. 

Thanks!

### DIFF
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -202,6 +202,7 @@ class HorizontalBarChart(graphics.Sprite):
         self.layout.set_markup("Hamster") # dummy
         self.label_height = self.layout.get_pixel_size()[1]
 
+        self._total = dt.timedelta()
         self._max = 0
 
     def set_values(self, values):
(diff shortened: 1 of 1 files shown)


src/hamster/overview.py
```diff
@@ -210,23 +211,31 @@ def set_values(self, values):
...
     def _draw(self, context, opacity, matrix):
         g = graphics.Graphics(context)
         g.save_context()
         g.translate(self.x, self.y)
 
+        hours = [(value.days * 24.0 + value.seconds / 3600.0 ) for _,value in self.values]
+        total = self._total.days * 24.0 + self._total.seconds / 3600.0
+
         for i, (label, value) in enumerate(self.values):
+            percent = 100.0 * hours[i] / total
```
HWiese1980: No error checking if `self._total` is not set (i.e. zero); in that case there will be a division by zero error here!

omron93:
+1 Works fine for me. Thanks.

elbenfreund:

For more context please see our [website](http://projecthamster.org).

ederag:
Thanks for your contribution.
I tried to merge that in august, while gathering fixes,
but the sums were not displayed, and the percentages took to much space.
This could be due to wrong merge choices on my part, or to changes in the UI.

The totals are displayed in the current master
(and already there in [v2.1.1](https://github.com/projecthamster/hamster/releases/tag/v2.1.1)).
Having the percentages as numbers would be great as well,
but it should be an option, since space is short.

To be investigated again after the [branch](https://github.com/ederag/hamster/tree/CendioOssman-layout2) described in PR #302 is merged,
because it would affect the same parts.

ederag:
That would be a good time to add the percentages,
since I'm going going to work on other parts.
It should be an option, since space is short.

matthijskooijman:
There is a pending check for CLA that is not actually valid. Seems we can't remove it, so just ignore it (see also #589).

PR #349, opened 2017-09-29
Added missing dependency for debian systems

NVolcz:
Hamster is unable to run on debian-based systems without the python-gi-cairo package. Added it to the dependency list in README.me.


ederag:
Sorry I overlooked this one and merged the similar but late PR #365 instead.
Thanks for contributing !

PR #363, opened 2018-08-23
python3 port + usability fixes

ederag:
hamster-time-tracker was dropped from openSUSE Leap15
https://www.reddit.com/r/openSUSE/comments/7zrj03/whats_the_best_time_tracker_on_tumbleweed/
The new development seemed to go well, but still claimed as alpha.
The [fork by gsobczyk](https://github.com/gsobczyk/hamster) had too many unresolved dependencies on Leap15.
The [master](https://github.com/projecthamster/hamster/commit/c3e5fb761c88fdecfd1566cac8b6836228a27cce) branch worked out of the box,
without python-gconf, and with the same hamster.db database.

Several usability fixes were needed though.
Now it seems more intuitive and consistent.
There is a python2 branch, but I'm now exclusively on the python3 branch.
There seems to be little chance to merge it back here,
so comments, issues and small PR are welcome in this [fork](https://github.com/ederag/hamster).
It will hopefully last only a few years, until the new hamster-* is stable.

PR #364, opened 2018-09-17
Fix Hamster window toggle on dbus method.

wojtex:
There's a bug in Hamster.

Hamster version: 1.04-3 (from Arch Linux)
Desktop Environment: i3

### Steps reproducing this bug

Run Hamster and close the Hamster window (e.g. by hitting Escape).
Then call DBus method `org.gnome.Hamster.Toggle`:
```bash
dbus-send --session --dest=org.gnome.Hamster --type=method_call --print-reply /org/gnome/Hamster org.gnome.Hamster.Toggle
```

### Expected result

As docstring says:

https://github.com/projecthamster/hamster/blob/a5bda5a59abfaa7bd98dc525e5e0a404598c6bcd/src/hamster-service#L129-L133

So Hamster window should show on the screen.

### Actual result

Hamster prints to the output:
```pytb
  File "/usr/lib/python2.7/site-packages/hamster/today.py", line 430, in on_toggle_called
    self.window.present()
AttributeError: 'NoneType' object has no attribute 'present'
```

So `self.window` is `None`.

### Solution

We should call `self.toggle_hamster_window()` instead of `self.window.present()`

### DIFF
--- a/src/hamster/today.py
+++ b/src/hamster/today.py
@@ -427,7 +427,7 @@ class DailyView(object):
 
 
     def on_toggle_called(self, client):
-        self.window.present()
+        self.toggle_hamster_window()
 
     def on_conf_changed(self, event, key, value):
         if key == "day_start_minutes":


ederag:
Thanks for the detailed PR.
I'm trusting you here, because the `hamster_stable` branch is [uninstallable on my system](https://github.com/projecthamster/hamster/pull/363).

Issue #76, opened 2013-02-05
Report total incorrect


ajrouvoet:
It's an early morning for me and I hope I haven't overlooked something incredibly stupid, but I think the report totals are incorrect for me.

Every month I export my monthly report to a tsv file and sum the whole lot for my employer. The grand totals always matched up, but since the latest update they don't anymore.
### Summed minutes in xls

3717 minutes = 61.95 hours
### Total time in hamster

62.4 hours

Any help is appreciated.

ederag:
This should be fixed by PR #429, thanks @GeraldJansen !

PR #380, opened 2019-01-22
add total duration in 'hamster list' command

madtibo:
Add a total time line at the bottom of the 'hamster list' command.

### DIFF
--- a/src/hamster-cli
+++ b/src/hamster-cli
@@ -332,14 +332,14 @@ class HamsterClient(object):
         print("-" * min(row_width, 80))
 
         cats = []
-        total_duration = 0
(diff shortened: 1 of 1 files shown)


ederag:
Thanks for your contribution.
It would be better to sum deltas (that are in seconds), 
to avoid rounding errors such as https://github.com/projecthamster/hamster/issues/76.

madtibo:
I saved the duration in seconds, is it better that way?

ederag:
What about c71aaf7226d243460e570f5b974688de7915bdc1 ?
Totals would use the same "human" format as in the duration column.

ederag:
Thanks for contributing !

Issue #381, opened 2019-01-26
overview not refreshed after a command line stop


ederag:
Overview is not refreshed after a command line stop.
With `Python 3.6.5` on openSUSE Leap 15.0.

To reproduce:
It should be fine, but if you are using `hamster-time-tracker`, make sure to backup your `hamster.db` first.
1) ```bash
   pkill -ef hamster
   src/hamster-service &
   src/hamster-windows-service &
   src/hamster-cli
   ```
1) create an ongoing fact in overview.
2) on another terminal, `src/hamster-cli stop`
The overview is not updated.

Here is snapshot 3245adbdbd85c7b4ee6c5ac93a0cf05132555968, showing the origin of this bug:
Despite the assignment, `self.facts` is still bound to the old `self.facts` list.

A forthcoming PR is hopefully solving it.

ederag:
The origin of this issue was perfectly explained by [this answer](https://stackoverflow.com/a/54400627/3565696).
This is a combination of the introduction of `Fact.__eq__` in #372 - precisely commit https://github.com/projecthamster/hamster/pull/372/commits/3837ddcc1b1ce5c36377ac9b355c6f92cb06dbdc,
https://github.com/projecthamster/hamster/blob/3837ddcc1b1ce5c36377ac9b355c6f92cb06dbdc/src/hamster/lib/__init__.py#L200-L202
and 
https://github.com/projecthamster/hamster/blob/3245adbdbd85c7b4ee6c5ac93a0cf05132555968/src/hamster/lib/graphics.py#L1861-L1863

PR #382, opened 2019-01-27
base Fact comparison on main attributes

ederag:
Seems to fix issue #381.

The `fact.id` is not modified when the `fact.end_time` changes,
but comparison of facts was based on `fact.id` alone,
and for some reason, this lead python to *not* rebind `self.facts` to the new `facts` after
https://github.com/projecthamster/hamster/blob/843f3638e62beff2f4b0145cdd362bc7b30cb0ee/src/hamster/widgets/facttree.py#L400
=> assigned but not identical. Weird. (further analysis in #381)

Note: `fact.id` is the hamster database id, not the python `id(fact)`.

With this PR, facts comparison is based on the main attributes.
(excluding `id` and `activity_id` attributes that are database implementation details)

### DIFF
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -28,8 +28,9 @@ import datetime as dt
 """
 from hamster import widgets
 from hamster.lib.configuration import runtime, conf, load_ui_file
-from hamster.lib.stuff import datetime_to_hamsterday
+from hamster.lib.stuff import hamster_today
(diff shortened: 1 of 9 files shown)


ederag:
#381 is fixed with this PR.
The last commit (7f527368d7b0debe2678c1ea3e021f85479c8be2) is a fix to #384
that used the `hamster_today` function.

PR #383, opened 2019-02-01
Fix comparison bug between date and datetime types in reports

webnard:


ederag:
Thanks for contributing.
That should not be needed since 8b166c475b3e69e5861851fb7aaa13c651f2db72.
What was your branch base commit ?
How to reproduce the bug ?

ederag:
Current master should be perfectly compatible with the old 1.04 version (back and forth).
Any incompatibility would be considered a bug.
What might have happened is a stray hamster-service running from the old version,
while the export was made through a newer version ?

ederag:
Should be fixed. Thanks again for contributing.
More tests will be welcome !

Issue #230, opened 2015-04-14
Keyboard shortcuts


decibyte:
From the overview window, it would be great to have some keyboard shortcuts for (suggestions in parentheses):
- New (Ctrl+N)
- Edit current (Ctrl+E)
- End current now (Ctrl+X)
- Focus search field (Ctrl+F)

In the edit window, Ctrl+Enter to save would be great  (although Alt+S works today), as this is sort of a de facto standard shortcut for clicking the OK button on things, at least on the web.

Support for optional global keyboard shortcuts for New, Edit, End would be even more useful.

nchachereau:
Hi there!
Hamster is currently ongoing a rather large transformation. With the new interface, Ctrl+N and Ctrl+F both works in the Overview window, as does "Enter" (or Ctrl+Enter) in the Activity window. Escape will close the windows.

There is still room for more, which is why I leave this bug open. I'm rather unconvinced by Ctrl+X to stop tracking, as it usually means "cut" and someone might try to "cut" a focused activity.

Note that Hamster does not get much attention, as the project's creator does not maintain it any more, and no contributor manages to spend much time on its development.

decibyte:
Ctrl+X was just a suggestion. The main point is that I'm missing some kbd shortcuts. Sounds good that some of it is already on the way. I'm looking forward to upcoming updates.

ederag:
Thanks for the inputs.

#387 improve shortcuts and overall keyboard handling.
Feedback would be welcome.

What should be the best default shortcut for stop.
`Ctrl-S` sounds natural, but it is very often used to save in other software.
`Ctrl-H` (Halt) ?

decibyte:
Ooh, lovely to see some attention here :)

It's a good question, as most of the obvious letters are taken. Maybe E (for "End")?

Just curious: Are you also working on global shortcuts?

ederag:
> Just curious: Are you also working on global shortcuts?

Global shortcuts can be done using a combination of the desktop environment and
the command-line interface (e.g. `hamster overview` to popup the overview window).
So in the far future, at best.
Unless someone comes up  with a clean solution that works on any desktop
(not just Gnome).

> Maybe E (for "End")?

It is really tough to find a non-ambiguous one;  E could mean "Edit" as well.
What about `Ctrl+Space` ?
or `Ctrl+P` for "Pause" ?
Those two could be used also to "Play/unPause" the last activity ?
(much later; design issues foreseen)

decibyte:
[...]

PR #387, opened 2019-02-09
Improve keyboard handling

ederag:
Shortcuts (#230)
Arrows to navigate between dates and facts.

### checks
The following are tests, and show potential uses.

Left/Right arrow: change dates

With nothing selected
Ctrl-R: do nothing
Ctrl-N: new empty fact
Ctrl-+: new empty fact
Ctrl-Enter or Enter: do nothing

Select one fact
Ctrl-R clone, starting now
Ctrl-N new empty fact
Ctrl-+ clone, starting now
Ctrl-Enter or Enter: edit selected fact

Create a new fact
if not selected
Delete: do nothing
select the fact
Delete: fact deleted

With enough dates to have more facts than can fit in the overview,
to check scrolling,
check that the following keys work also right at the start of overview,
(without having to focus on the FactTree itself)
Up, Down,
Home, End (change selected fact)
Page up, Page down, (just scroll)

With an ongoing fact
Ctrl+Space: stop

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,19 @@
 ===============
+Changes since v2.2.2
+===============
+* Improved keyboard handling:
+  - Ctrl+Space to stop tracking.
+  - Left/Right arrows change date.
+  - Resume: start now a clone of the selected activity.
+    Ctrl-+: clone or fallback to new if none selected.
+            (same as pressing the + button)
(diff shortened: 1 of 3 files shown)


GeraldJansen:
Nice! Thanks.

PR #393, opened 2019-03-01
Fix some install paths, and provide help menu entry

mwilck:
These fixes were created while I was working on the openSUSE package lately. Well, I guess the second one is more a feature addition than a bug fix, although I like it a lot. I can create another PR without it if you prefer.

### DIFF
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -557,7 +557,7 @@ class Overview(Controller):
                 webbrowser.open_new("file://%s" % path)
             else:
                 try:
-                    gtk.show_uri(gdk.Screen(), "file://%s" % os.path.split(path)[0], 0)
+                    gtk.show_uri(None, "file://%s" % path, gdk.CURRENT_TIME)
                 except:
                     pass # bug 626656 - no use in capturing this one i think
 
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -90,6 +90,8 @@ class HeaderBar(gtk.HeaderBar):
         self.system_menu.append(self.menu_export)
         self.menu_prefs = gtk.MenuItem(label=_("Tracking Settings"))
         self.system_menu.append(self.menu_prefs)
+        self.menu_help = gtk.MenuItem(label=_("Help"))
+        self.system_menu.append(self.menu_help)
         self.system_menu.show_all()
 
 
@@ -434,6 +436,7 @@ class Overview(Controller):
 
         self.header_bar.menu_prefs.connect("activate", self.on_prefs_clicked)
         self.header_bar.menu_export.connect("activate", self.on_export_clicked)
+        self.header_bar.menu_help.connect("activate", self.on_help_clicked)
 
 
         self.window.connect("key-press-event", self.on_key_press)
@@ -537,6 +540,9 @@ class Overview(Controller):
         self.find_facts()
         # The timeout will stop if returning False
         return True
+
+    def on_help_clicked(self, menu):
+        gtk.show_uri(None, "help:hamster-time-tracker", gdk.CURRENT_TIME)
 
     def on_prefs_clicked(self, menu):
         dialogs.prefs.show(self)
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Project Hamster.  If not, see <http://www.gnu.org/licenses/>.
 
+import sys
 import bisect
 import datetime as dt
 import itertools
@@ -542,7 +543,18 @@ class Overview(Controller):
         return True
 
     def on_help_clicked(self, menu):
-        gtk.show_uri(None, "help:hamster-time-tracker", gdk.CURRENT_TIME)
+        uri = "help:hamster-time-tracker"
+        try:
+            gtk.show_uri(None, uri, gdk.CURRENT_TIME)
+        except gi.repository.GLib.Error:
+            msg = sys.exc_info()[1].args[0]
+            dialog = gtk.MessageDialog(self.window, 0, gtk.MessageType.ERROR,
+                                       gtk.ButtonsType.CLOSE,
+                                       _("Failed to open {}").format(uri))
+            fmt = _('Error: "{}" - is a help browser installed on this computer?')
+            dialog.format_secondary_text(fmt.format(msg))
+            dialog.run()
+            dialog.destroy()
 
     def on_prefs_clicked(self, menu):
         dialogs.prefs.show(self)
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,27 @@
 ===============
+Version 2.2.1
+===============
+* Activity can contain decimal numbers.
+* Add total duration to the `hamster list` output.
+* Base Fact comparison on main attributes.
+* Small improvements to suggestions.
+* A selected fact can be resumed (start now a clone)
+  by pressing the `+` button.
+* Facts can be deselected by mouse click.
+* Still ongoing activities are shown,
+  even if they started on a previous day.
+* Several export (reports) fixes.
+* Fact.delta (duration of the fact, as a timedelta)
+  is a read-only property.
+* Tag list keeps the input order.
+* Better format of the stats area.
+
+Thanks to all contributors (either code or feedback) !
+More information by looking for milestone:2.2 or by
+cloning and using the following command:
+gitk v2.1.1...v2.2.1
+
+===============
 Version 2.1.1
 ===============
 Migration to python3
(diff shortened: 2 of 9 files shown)


ederag:
Sounds interesting, thanks !
I would like to merge the help stuff as a whole so yes,
please move out the bash-completion commit to another PR, since it is unrelated.
There will be some delay, as I'm going to set up a virtualbox
to ensure that installation do not break on ubuntu.

ederag:
In `lubuntu-18.10` for the development version, 
when the help files are not installed 
```python
  File "/media/sf_python/hamster/src/hamster/overview.py", line 546, in on_help_clicked
    gdk.CURRENT_TIME)
gi.repository.GLib.Error: g-io-error-quark: The specified location is not supported (15)
```
Having an error is normal (the help files are missing), 
but the error message could be clearer.

By the way, the help files are not installed by 
```bash
./waf configure build --prefix=/usr && sudo ./waf install
```
even when all dependencies are satisfied :
`python3-gi-cairo gir1.2-gconf-2.0 python3-xdg python3-dbus gettext intltool gnome-doc-utils docbook-utils`
How should the help files be installed ?

mwilck:
> In `lubuntu-18.10` for the development version,
> when the help files are not installed
> 
> ```python
> Traceback (most recent call last):
>   File "/media/sf_python/hamster/src/hamster/overview.py", line 546, in on_help_clicked
>     gdk.CURRENT_TIME)
> gi.repository.GLib.Error: g-io-error-quark: The specified location is not supported (15)

Do you have the GNOME help browser installed (yelp)? It's responsible to display help files.

But yeah, I guess we need to catch that error, similar to how it's done for in the "export" code when opening non-HTML files. Althougth the generic "except: pass" there looks a bit too catch-all to my taste.

> By the way, the help files are not installed by
> 
> ```shell
> ./waf configure build --prefix=/usr && sudo ./waf install
> How should the help files be installed ?

I was looking at the "waf configure" output which complained about some tools being missing. See [the build log](https://build.opensuse.org/build/devel:openSUSE:Factory:Apps/openSUSE_Tumbleweed/x86_64/hamster-time-tracker/_log). I added xml2pod (and xsltproc, but I don't think it's necessary), which caused the help files to be generated and installed. Waf also looks for docbook2man, but AFAICS it doesn't do anything with it, and docbook2man (at least on SUSE) pulls in an enormous list of build dependencies via tetex, so I did not add it.

mwilck:
> Do you have the GNOME help browser installed (yelp)? It's responsible to display help files.

Well it's "yelp" under GNOME. Other desktops may have their own help display tools. But the install location and the file format are cross-desktop AFAICT.

mwilck:
I pushed another patch to handle the "help browser not available" problem, and dropped the bash-completion thing for now.

src/hamster/overview.py
```diff
@@ -538,6 +542,21 @@ def on_timeout(self):
         # The timeout will stop if returning False
         return True
 
+    def on_help_clicked(self, menu):
+        uri = "help:hamster-time-tracker"
+        try:
+            gtk.show_uri(gdk.Screen().get_default(), uri,
```
ederag:
```suggestion
            gtk.show_uri(None, uri,
```
`None` seems the standard way to say "default location"
https://lazka.github.io/pgi-docs/#Gtk-3.0/functions.html#Gtk.show_uri
https://programtalk.com/python-examples/gi.repository.Gtk.show_uri/
It does not seem possible to make a suggestion on several lines,
but of course it should be `gtk.show_uri(None, uri, gdk.CURRENT_TIME)`

src/hamster/overview.py
```diff
@@ -538,6 +542,21 @@ def on_timeout(self):
...
+                         gdk.CURRENT_TIME)
+        except gi.repository.GLib.Error:
+            msg = sys.exc_info()[1].args[0]
+            dialog = gtk.MessageDialog(self.window, 0, gtk.MessageType.ERROR,
+                                       gtk.ButtonsType.CLOSE,
+                                       _("Failed to open {}").format(uri))
+            fmt = _('Error: "{}" - is a help browser installed on this Computer?')
```
ederag:
```suggestion
            fmt = _('Error: "{}" - is a help browser such as yelp installed on this Computer?')
```
Mentioning `yelp` to further guide the user ?

src/hamster/overview.py
```diff
@@ -557,7 +576,8 @@ def on_report_chosen(widget, format, path):
                 webbrowser.open_new("file://%s" % path)
             else:
                 try:
-                    gtk.show_uri(gdk.Screen(), "file://%s" % os.path.split(path)[0], 0)
+                    gtk.show_uri(gdk.Screen().get_default(),
```
ederag:
```suggestion
                    gtk.show_uri(None,
```
As above,
`None` seems the standard way to say "default location"
https://lazka.github.io/pgi-docs/#Gtk-3.0/functions.html#Gtk.show_uri
https://programtalk.com/python-examples/gi.repository.Gtk.show_uri/
It does not seem possible to make a suggestion on several lines,
but of course it should be `gtk.show_uri(None, "file://%s" % path, gdk.CURRENT_TIME)`

ederag:
> How should the help files be installed ?

Good news: there was only a cache issue. The safest build is
```bash
./waf clean && ./waf distclean && ./waf configure build --prefix=/usr && sudo ./waf install
```
(both `clean` and `distclean` were necessary to get the help built)

src/hamster/overview.py
mwilck: I don't think `yelp` is the only possibility. The KDE Help browser is another I guess. I wouldn't want to be too specific, or biased towards a certain environment. Next year "yelp" may have been replaced by something else in GNOME.

src/hamster/overview.py
mwilck: Yes, "None" works, too. It comes down to exactly the same thing AFAICT.

src/hamster/overview.py
mwilck: Moreover, the name "yelp" is also pretty much hidden in GNOME environments. I believe that using such specific names is pretty much deprecated in GUI environments. "Help browser" should mean something to most users, "yelp" definitely doesn't. It's true that users need to figure out which package they need to install to get the help browser - but honestly that's GNOMEs or distributor's job, not ours.

src/hamster/overview.py
mwilck: If you read that documentation, you'd expect Gdk.Screen() to be correct as first arg to Gtk.show_uri(). But it crashes reliably, I tested that in various enviroments. I wonder if we have spotted a problem in PyGTK (or GI, for that matter). But I don't care that much, actually.

mwilck: The types of gdk.Screen() and gdk.Screen.get_default() are different:

    >>> gdk.Screen()
    <Gdk.Screen object at 0x7f7e89bd4f00 (GdkScreen at 0x56502c04eb60)>
    >>> gdk.Screen().get_default()
    <GdkX11.X11Screen object at 0x7f7e89b8d690 (GdkX11Screen at 0x56502bf36000)>

mwilck:
I made the "None" change now. Feel free to modify the error message as you see fit.

src/hamster/overview.py
ederag: That makes sense. We'll just add a line in the [dependencies section](https://github.com/projecthamster/hamster#dependencies).

mwilck: Make sure to make it a "weak" dependency though. In distro speak, it would be a "Recommends:", not a "Requires:".

ederag: The KDE help browser `khelpcenter` does not seem to read these `mallard` pages.
(they do not show up when searching for "hamster").
Yelp seems to be the only solution currently. No big deal; it is both nice and lightweight.
https://github.com/mwilck/hamster/pull/3 starts with a proposition about that.

src/hamster/overview.py
ederag: Interesting, but the `None` version you pushed seems fine, thanks.

ederag:
Comments left above; let's continue in https://github.com/mwilck/hamster/pull/3 for now.

mwilck:
Merged your suggestion, your turn now ;-)

ederag:
Well thanks, but I was not done :smile: 
The help pages are very good, but still there is plenty of work to update them first
(broken links, some out-of-date stuff, ...). 
It will take a while (hence the work in progress header in my PR).

mwilck:
> The help pages are very good, but still there is plenty of work to update them first

Alright ... I thought that could be done in a later step. But fine with me this way.

ederag:
Thanks for your comprehension.
The help pages are very important, and this will make a nice homogeneous bunch of commits.

ederag:
@mwilck Thanks for bringing the help back to life !

PR #404, opened 2019-04-23
Use GLib.MainLoop instead of GObject.MainLoop

GeraldJansen:
Fix deprecation message: hamster/src/hamster-service:14: PyGIDeprecationWarning: GObject.MainLoop is deprecated; use GLib.MainLoop instead.

### DIFF
--- a/src/hamster-service
+++ b/src/hamster-service
@@ -3,6 +3,7 @@
 
 import logging
 from gi.repository import GObject as gobject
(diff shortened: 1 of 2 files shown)


ederag:
Looks good.
And following to [this discussion](https://github.com/matplotlib/matplotlib/pull/11182/), sounds safe.
Thanks !

PR #408, opened 2019-04-24
remove unused "trophies" code

GeraldJansen:
Tidy up a bit by removing no-longer-used code for "trophies" (missing module gnome_achievements).


GeraldJansen:
Thanks!

PR #407, opened 2019-04-23
improve docs on merging/splitting activities

GeraldJansen:
Clarify how overlapping activities are split/merged. Refs #396.

Issue #396, opened 2019-03-07
Split activity


ederag:
According to the help pages,
 https://github.com/projecthamster/hamster/blob/8326504b4a96b4cbc798b67efec27b3fdbbe83d1/help/C/merge.page#L11-L14

This does not work this way now. (the end part is simply disregarded)
The relevant code is still there:
https://github.com/projecthamster/hamster/blob/8326504b4a96b4cbc798b67efec27b3fdbbe83d1/src/hamster/storage/db.py#L454-L456
Need to investigate...

GeraldJansen:
Hmm, works as documented when the activity being split is not the last of the day. For example, I inserted 3 entries, 1) "0900 1200 activity1", 2) "1400 activity2" (in progress) and 3) "1000 1100 last activity" and I ended up with:
![Screenshot_2019-04-23_17-48-19](https://user-images.githubusercontent.com/31414956/56596071-20e5d380-65f0-11e9-9f1a-83bb9a0dd492.png)

GeraldJansen:
Upon further testing, I find that it always works as advertised, but the documentation could be clearer. When you create a new activity whose start  and end time both lie within the interval of an existing activity, the existing activity will be split in two. If you only specify a start time then the previous activity is simply truncated, not split.

ederag:
interesting observation, thanks !

GeraldJansen:
Whether it fails or is a feature is a matter of opinion I suppose. Currently the behaviour is to truncate the ongoing activity at the start of the new activity, whether or not the new activity has an end time. It is done in __add_fact [here](https://github.com/projecthamster/hamster/blob/504220b81c12b27f20f3370d2863562292b68ffa/src/hamster/storage/db.py#L600).

ederag:
In my opinion, it would more consistent to split the ongoing activity A in two, inserting the new B inside,
and leaving the last part ongoing (yielding A B A-ongoing).

Besides, trying to insert a B without any end-time yields two on-going activities,
instead of making A end when the ongoing B starts.

But that will have to wait, after 2.3 to the very least.

GeraldJansen:
I agree about postponing this work.

> Besides, trying to insert a B without any end-time yields two on-going activities, instead of making A end when the ongoing B starts.
I cannot replicate this behaviour. For me, ongoing activity A is terminated at the start-time of B, whether or not B has an end-time. 

[...]

PR #410, opened 2019-04-25
Fix an overlapping activity bug

### LINKED ISSUE #409: Failure on special case of overlapping activities
Hamster fails when inserting a new activity that spans the start time of an ongoing activity. In grafical mode the "Add earlier activity" window does not respond when you try to save the activity. In development mode:
```
[src]$ ./hamster-cli list
Start | End | Duration | Activity | Category
-----------------------------------------------
08:36 |     | 0h 5min  | xxx      | Unsorted
-----------------------------------------------

[src]$ ./hamster-cli track yyy 08:30 08:40
  File "./hamster-cli", line 414, in <module>
  ...
  File "/d/gj/hamster/src/hamster/storage/db.py", line 483, in __solve_overlaps
    if start_time < fact["start_time"] and end_time > fact["end_time"]:
TypeError: '>' not supported between instances of 'datetime.datetime' and 'NoneType'
```

GeraldJansen:
Fix #409.

### DIFF
--- a/src/hamster/storage/db.py
+++ b/src/hamster/storage/db.py
@@ -479,13 +479,15 @@ class Storage(storage.Storage):
                                           start_time, end_time))
 
         for fact in conflicts:
+            fact_end_time = fact["end_time"] or dt.datetime.now()
+
             # won't eliminate as it is better to have overlapping entries than loosing data
-            if start_time < fact["start_time"] and end_time > fact["end_time"]:
+            if start_time < fact["start_time"] and end_time > fact_end_time:
                 continue
 
             # split - truncate until beginning of new entry and create new activity for end
-            if fact["start_time"] < start_time < fact["end_time"] and \
-               fact["start_time"] < end_time < fact["end_time"]:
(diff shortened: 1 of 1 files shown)


ederag:
Interesting. After a first test, it does seem to fix #409.
But as said in a [previous comment](https://github.com/projecthamster/hamster/pull/407#issuecomment-485949845), to me touching this code is really scary at this time.

But if that is important to you, I will find the time and energy to get into that.

GeraldJansen:
Neither this isolated bug nor a general review of splitting/merging is that important to me. I came upon this bug while testing bigger changes to this section of code, trying to get a better understanding of #396. I tried to isolate and minimize the changes and did my best to test that other cases of overlap were not affected. I feel reasonably confident that these changes will not break current behaviour, and even if they do it will be very easy to back out. If you prefer to wait post v2.3, then fine by me.

ederag:
Indeed, I also find these changes sound and cautious.
So I looked a bit deeper into `__add_fact` and `__solve_overlaps`,
and I'm also relatively confident that this is a safe improvement.
Thanks !

Issue #274, opened 2015-12-16
Task description size


thinkdao:
Task description takes overflow

![description_bug](https://cloud.githubusercontent.com/assets/16149867/11853908/8bacf0e0-a438-11e5-9bab-e4b79f522b98.jpeg)

ederag:
Should be solved by PR #412.
Thanks for the clear report !

PR #412, opened 2019-05-04
Fix multiline description height

ederag:
The height must be calculated on the fly, 
otherwise multiline labels would still have a single line height.
Solves #274.


ederag:
0eb2c239c154f496139c7e53e262e947512e2c3a fixed #413 actually.

Issue #413, opened 2019-05-11
Incorrect highlighting of multiline descriptions


GeraldJansen:
Multiline descriptions (with embedded newlines) are not properly handled in the overview window. As shown in the screenshots below, when selecting tasks with the up/down arrows, the first task gets highlighted with the size of the second task and vice-versa. The selection of tasks by clicking on them is also messed up because of incorrect height calculations. This may be related to #274 and #412. 
(For me this is low priority because I don't personally use descriptions, but I thought it worthwhile  to document the issue anyway.)

![Screenshot_2019-05-11_16-59-14](https://user-images.githubusercontent.com/31414956/57571959-fd56c180-7414-11e9-8a16-a6ffec196b94.png)

![Screenshot_2019-05-11_16-57-55](https://user-images.githubusercontent.com/31414956/57571968-1bbcbd00-7415-11e9-96a5-a188ec99d939.png)

ederag:
Good observation, thanks ! I'm on it.

ederag:
Should be fixed by PR #414

PR #414, opened 2019-05-15
Fix multiline descriptions

ederag:
Solves  #413, that was actually hidden before #412.
By the way, fixed row height to avoid tag spill-out such as
image

Also fixed the category color. 
Now it has always a lower contrast, even when the row is selected.
And that works for other themes (tested with dark adwaita).
Under adwaita the new color is within 5% of the old one.

Developer-wise, the margins are defined in variables, instead of hard-coded.
Changing one of these variables value should still give consistent results.


GeraldJansen:
For what it's worth I confirm that these changes fix the multiline description problems. In my limited testing I haven't noticed any new problems. Thanks.

ederag:
Thanks for the feedback !

PR #415, opened 2019-05-19
RFC: preselect time instead of activity

ederag:
Most often, it is the time that has to be changed (at least for me).
This is especially true when cloning.

This changes the initial selection in the activity entry line
to be the time information.

Selecting the body can be done quickly by pressing `Right arrow` then  `Shift+End`.

Is there a a consensus about this, or do we have to add an option ?


GeraldJansen:
I have tested the change and like it.

ederag:
Thanks for the feedback !
Unless someone chimes in, I'll merge that this week-end.

Issue #399, opened 2019-03-16
After waf ?


ederag:
There are issues with the current installation system (`waf`): #309, #221, #252, ...

The current well established standard is [`setuptools`](https://setuptools.readthedocs.io/en/latest/setuptools.html). We could use the[`setup.cfg`](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) configuration file.
[Here](https://github.com/ederag/geoptics/blob/master/setup.cfg) is an example.
This seems to be the safest course (stability is important for this repository).

matthijskooijman:
I'm surveying what the current waf script does exactly, other than building and installing the python module. I've identified:
 - Install a dbus service file (in order to autostart the hamster daemon)
 - Install top-level service script (in `/usr/lib/hamster-time-tracker` I believe) for dbus to start
 - Install bash completion files
 - Postprocess gconf schemas using `gconftool-2 --install-schema-file` (schemas are installed by `data/wscript`).
 - Postprocess icons using `gtk-update-icon-cache -q -f -t` (icons are installed by `data/wscript`).
 - (through `po/wscript`) install translations (I think)
 - (through `data/wscript`) install various data files (images, gtk ui files, hamster.db, report template) into `/usr/share/hamster-time-tracker`.
 - (through `data/wscript`) install icons into `/usr/share/hamster-time-tracker`.
 - (through `data/wscript`) translate and install gconf schemas into `/etc/gconf/schemas` (I *think*).
 - (through `data/wscript`) translate and install `.desktop` files into `/usr/share/applications`.
 - (through `help/wscript`) install help files into (I think) `/usr/share/gnome/help/hamster-time-tracker` (and some more stuff I didn't immediately understand)

This is a lot more than a default python package usually does. Before making this list, I was hoping to limit what needs to happen so we could be more like a plain python package that can just be pip installed (e.g. by removing the dbus autostart, which [might be a good idea anyway](https://jdebp.eu/Softwares/nosh/avoid-dbus-bus-activation.html), but seeing this list I doubt that that is actually possible  (and we'd have to rely on OS packages anyway). That does not mean we cannot use setuptools or something similar, though (I expect setuptools or other build systems can  be made to do the above as well).

[...]

PR #421, opened 2019-07-11
Waf update

ederag:
This is a PR for the waf update (issue #399).

Need to (at least):
- [x] add a license file to the waflib directory
- [x] restore icon cache updates
- [x] new waf dbus needs `dbus-binding-tool` ?
- [x] installation instructions: documentation needs `yelp`
- [x] restore help build
- [x] make help installation optional (wip + need doc)
- [x] update scrollkeeper ?
- [x] change default prefix to `usr`
- [x] do not use `conf.define` (get rid of `compat15`)
- [x] more tests + instructions for other distros
- [x] add uninstall instructions
- [x] update `NEWS`


ederag:
Docs and main app have been separated.
`hamster` works fine on ubuntu-19.04. The help should be available.
Please note that installation instructions changed.

ederag:
https://github.com/projecthamster/hamster/pull/421/commits/66baf834540018046aef384d925273d48eee7be5 looks like a hack,
and splitting the documentation and main app builds brings trouble for the uninstall process.
Let us just skip the documentation build if xml2po is not found, to reduce dependencies.

ederag:
Done.
Tested on clean openSUSE Leap 15.1, ubuntu 18.04 and 19.04 virtualboxes.

Feedback would be  welcome !

For review, here are the [changes except the `waf` update itself](https://github.com/projecthamster/hamster/pull/421/files/e265292c05f7782b82895e81d912115d89508496..0a23d68ee18cf5716b810c372b2ccbd57d46294b).

GeraldJansen:
> Feedback would be welcome !

Works for me on Xubuntu 19.04! Including the help pages. Many thanks for all that work!

The only glitch I ran into is that I normally use `umask 0007` and that caused the files installed in /usr to be not readable by regular users. After uninstalling, and starting over with `umask 0002` all went smoothly.

Maybe there is not much feedback because some people (like me) don't know how to go about testing a pull request. This is the way I did it:
`git clone --branch=waf-update git@github.com:ederag/hamster.git hamster-waf-update`
but there are probably better ways (git add remote?).

Before installing I did a `sudo ./waf uninstall` from my previous version and `sudo rm -rf /usr/lib/python3/dist-packages/hamster`. I also found a few other leftover files with `find /usr -name "hamster*"` and removed those too.

ederag:
Thanks for the feedback !

> Maybe waf could set the umask to avoid this issue.

Forcing umask might be disturbing.
`pip` [use standard umask](https://stackoverflow.com/a/11169137/3565696), and so does `waf`.
A warning would be nice, but how to get the default umask ?

> [...] there are probably better ways (git add remote?).

Thanks for the "one-shot" command !
Wiki [updated](https://github.com/projecthamster/hamster/wiki/How-to-contribute/_compare/946c7f363fdd5d5bddd5a7ddb5544d1513fc783f...035b5721acda4321a6930fe8599775975819aafc) with more complete instructions.

> I also found a few other leftover files with `find /usr -name "hamster*"`

That should not happen in principle. Here is a guess.
To correctly remove a version earlier than this PR (e.g. `v2.2.2`), 
the `--prefix=/usr` is required for the configure step:
```bash
git checkout v2.2.2
./waf configure --prefix=/usr
sudo ./waf uninstall
```

GeraldJansen:
Thanks for the tip about umask and `sudo pip install`: finally I know why I've always had those permission problems installing with pip! For hamster, maybe just change the docs to `( umask 0022 && sudo ./waf install; )`. Works for me and leaves my usual umask unaltered.

My "hamster*" left-overs in /usr may have come from some manual interventions trying to get the docs/yelp working with xfce in various past versions. Uninstall works cleanly now.

Thanks also for the improvements in the documentation about Contributing. Super.

ederag:
> `( umask 0022 && sudo ./waf install; )`. Works for me and leaves my usual umask unaltered.

That is quite surprising, as `sudo` is designed to be insensitive to the calling environment.
Indeed here, 
```bash
> ( umask 0002 && sudo bash -c umask; )
0022
```
Trying to understand, I found this very [interesting answer](https://askubuntu.com/a/276958/821405) about default umask in ubuntu.

GeraldJansen:
Based on [this](https://unix.stackexchange.com/questions/265497/set-sudo-umask-apart-from-the-user-calling-it), sudo usually is sensitive to the user's umask. My solution is not very elegant but i protects against an overly restrictive umask. Now I've added these lines to /etc/sudoers so I personally won't have the problem anymore:
    `Defaults umask=0022`
    `Defaults umask_override`

ederag:
Thanks for the link, this makes sense now.
Something has to be done. I would add some explanations, 
otherwise that might look suspicious to people who never faced the issue.
For instance:
"""
```bash
# thanks to the parentheses the umask of your shell will not be changed
( umask 0022 && sudo ./waf install; )
```
The `umask 0022` is safe for all, but important for users with more restrictive umask,
as discussed [here](https://github.com/projecthamster/hamster/pull/421#issuecomment-520167143).
"""

Finally, it happens that the instructions are not exactly copy/pasted anyway.
So to help finding this discussion, in case of a too restrictive umask, launching `hamster` yields
```python
  File "/usr/bin/hamster", line 30, in <module>
    from hamster import client, reports
ImportError: cannot import name 'client' from 'hamster' (unknown location)
```

ederag:
Thanks again for the feedback Gerald !

GeraldJansen:
A big step. Yahoo!

Issue #312, opened 2016-03-28
Activities text too small on HiDPI screens


AdrianSimionov:
Activities tab has the text too small, it does not seem to respect the DPI of the system.

![march 27 april 02 2016_630](https://cloud.githubusercontent.com/assets/6671693/14071655/f48d00ee-f502-11e5-926b-7ac024f68922.png)

karolszk:
Hello, I also confirm that on Debian Stretch on resolution 2560x1440 (16:9) activities are too small. See my screenshot.

chri2:
Changed it for totals on my system for the same reason: HiDPI. Look [here](https://github.com/projecthamster/hamster/issues/289). Maybe it helps to change this tab also...

elbenfreund:
Thank you for your feedback, it is much appreciated.

Project hamster and its various sub-components is currently undergoing some major changes. We prepare the introduction of a rewritten codebase for most of the underlying functionality. A direct consequence of this is that it is unlikely that any open/new bugs within the current/old codebase will be fixed (unless someone steps up and offers to do so) as most resources currently available will be invested in making the rewrite prime time ready.

ederag:
Thanks for the reports.
This issue seems related to the old hamster-applet (1.0.4 ?)
This seems fixed now, probably thanks to the v2 gtk3 migration
(tested on openSUSE Leap-15, 2560x1440, [v2.1.1](https://github.com/projecthamster/hamster/releases/tag/v2.1.1))

GeraldJansen:
Not sure if this is the same problem, but even with current master on XFCE, some fonts reflect the custom DPI settings and others do not.
![Screenshot_2019-08-11_16-29-00](https://user-images.githubusercontent.com/31414956/62835225-06467180-bc56-11e9-85de-f9d3a412be1f.png)

ederag:
Fixed by @GeraldJansen in PR #426, thanks !
Leaving open, but with the `hamster-applet` label, since it is still an issue in version 1.

GeraldJansen:
I feel that we should close issues in hamster-applet that have been fixed in current hamster master, after providing a kindly explanation. That software is now ancient and realistically no-one is ever going to work on it again. Excessive open issues give the impression that the software is buggier and less complete than it actually is. Personally I find it demotivating to fix an issue and not see it closed. 
(Btw, I guess #289 is the same issue)

ederag:
I hear you, thanks for sharing your thoughts.
[...]

Issue #289, opened 2016-02-13
Font#size on High DPI screens doesn't obey overall zoom in Unity


rubo77:
On Ubuntu on my Hi-res screen some fonts are resized and some are unreadeable small

I use scale factor x2

chri2:
Had the same problem. Even though I don't know python, worked my way through the code and found the place where fixed fonts are used for totals:

`diff /usr/lib/python2.7/dist-packages/hamster/lib/charting.py.orig /usr/lib/python2.7/dist-packages/hamster/lib/charting.py`
```
37,38c37,38
<         self.label = graphics.Label(value, size=8, color=label_color)
<         self.label_background = graphics.Rectangle(self.label.width + 4, self.label.height + 4, 4, visible=False)
---
>         self.label = graphics.Label(value, size=14, color=label_color)
>         self.label_background = graphics.Rectangle(self.label.width + 7, self.label.height + 7, 7, visible=False)
56c56,57
<         horiz_offset = min(10, self.label.y * 2)
---
>         # horiz_offset = min(10, self.label.y * 2)
>         horiz_offset = 5 
141c142
<             label = graphics.Label(stuff.escape_pango(key), size = 8, alignment = pango.ALIGN_RIGHT)
---
>             label = graphics.Label(stuff.escape_pango(key), size = 14, alignment = pango.ALIGN_RIGHT)
167c168
<         self.plot_area.x = legend_width + 8
---
>         self.plot_area.x = legend_width + 14 
241c242
<         self.graph_x += 8 #add another 8 pixes of padding
---
>         self.graph_x += 14 #add another 8 pixes of padding
248c249
<         default_font.set_size(8 * pango.SCALE)
---
>         default_font.set_size(14 * pango.SCALE)```

In my setup the font in this place at least is readable.

elbenfreund:
Thank you for your feedback, it is much appreciated.

Project hamster and its various sub-components is currently undergoing some major changes. We prepare the introduction of a rewritten codebase for most of the underlying functionality. A direct consequence of this is that it is unlikely that any open/new bugs within the current/old codebase will be fixed (unless someone steps up and offers to do so) as most resources currently available will be invested in making the rewrite prime time ready.

ederag:
Thanks for the investigation.
This issue seems related to the old hamster-applet (1.0.4 ?)
This seems fixed now, probably thanks to the v2 gtk3 migration
[...]

PR #426, opened 2019-08-12
Fix too small fonts on high DPI screens

GeraldJansen:
The use of pango.SCALE to set font sizes results in fonts that are too small on high DPI screens (see #312 and #289). This pull request attempts to resolve the issue by only using relative font sizes (i.e. &lt;small&gt; tag) in the fact tree display.

### DIFF
--- a/src/hamster/widgets/facttree.py
+++ b/src/hamster/widgets/facttree.py
@@ -22,7 +22,6 @@ import cairo
 import datetime as dt
 
 from collections import defaultdict
-
 from gi.repository import GObject as gobject
 from gi.repository import Gtk as gtk
 from gi.repository import Gdk as gdk
@@ -31,7 +30,6 @@ from gi.repository import Pango as pango
 
 from hamster.lib import graphics
 from hamster.lib import stuff
-
 
 
 class ActionRow(graphics.Sprite):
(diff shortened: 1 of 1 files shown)


GeraldJansen:
With this pull request:
Screenshot_2019-08-12_18-48-16

ederag:
`<small>{}</small>` looks like the smart and correct way to handle this, thank you so much !

Another way can be found in the no longer used `tags.py` (probably the fancy tags of version 1):
https://github.com/projecthamster/hamster/blob/d864d0a9ae872db497eb310ad67929d4f169dd6c/src/hamster/widgets/tags.py#L328-L329
but I much prefer using the pango designer skills with `<small>` tag, indeed.

Maybe we should use `stuff.escape_pango(text)` on the tag text.
I can file a separate PR for that if you prefer ?

While producing PR #414, I also have considered removing the `category_offset_V` downshift,
and eventually became quite convinced that it does help 
visually separating the category from the activity name.
We could make this downshift proportional to the category label height.
Or do you think that the downshift is annoying ?

GeraldJansen:
I thought the category_offset_V was related to the larger font size for category and that it didn't make much sense to keep it if the same font size is used for activity and category. Personally I find that the category is nicely highlighted and no further differentiation is needed. Here we get into personal tastes and I'm no designer!

ederag:
There is another `self.tag_label.set_text( "<small>{}</small>".format(fact.tags[0]))`.
To avoid code duplication, we could make a `class TagLabel(Label)` overloading `set_text`.
I'm going to push propositions for that and for the downshift to your branch.

GeraldJansen:
> There is another `self.tag_label.set_text( "<small>{}</small>".format(fact.tags[0]))`.
If I understand the code, that one is only used to determine row height but not displayed, so there is no need to escape it. Anyway, won't do harm I suppose.

ederag:
You are right of course ! Anyway, as you say it does not hurt, and it was almost done.
Here is the before / after restoring the downshift.
Screenshot_20190814_142030 Screenshot_20190814_142230
The downshift adds a little help, in addition to the contrast,
and I would tend to trust the design skills of our predecessors again.

GeraldJansen:
Okay for me.

ederag:
Just noticed that descriptions were not in italic. That would be restored by the previous commit.
What do you think ?

GeraldJansen:
Whatever.

ederag:
Merging then. Thank you so much !
I did not see how to fix that.

Issue #258, opened 2015-10-06
Can't stop an activity from overview window


sixtyfive:
So I rejoiced at a working Hamster too soon. Just tried to stop an activity for the first time and discovered that there's no way (no button, no context menu item) to do so. Do I need to compile a different branch instead of master or should this work somehow?

toupeira:
Ah yes this isn't exactly too obvious: in the main window you can stop an activity by either entering an end time (i.e. change "14:00 foo" to "14:00 16:00 foo") or by starting another activity. You can also use the CLI command `hamster stop`. But really Hamster is meant to be used through an [extension like the one for Gnome-Shell](https://github.com/projecthamster/shell-extension) which includes a "Stop activity" entry in the dropdown menu.

I'm keeping this issue open though, a context menu in the main window would make a lot of sense.

sixtyfive:
Thank you for the explanation!

Indeed, I would use Cinnamon's applet, but:

```
error t=2015-10-07T20:16:07.012Z [Applet "hamster@projecthamster.wordpress.com"]: Error importing applet.js from hamster@projecthamster.wordpress.com
error t=2015-10-07T20:16:07.013Z Could not load applet hamster@projecthamster.wordpress.com
```

toupeira:
@sixtyfive sorry I'm not familiar with the Cinnamon extension, but it looks like you're hitting https://github.com/projecthamster/cinnamon-applet/issues/14

ederag:
It is now [easier to stop an activity](https://github.com/projecthamster/hamster/issues/310#issuecomment-455236628)
but this could be improved; the `ActionRow` seems very promising:
https://github.com/projecthamster/hamster/blob/83edea9222acae426f8285999744f9462a2bdd03/src/hamster/widgets/facttree.py#L228-L229
(Not visible because commented out)

ederag:
With PR #387 and specifically 5c321307b0ecf5e67295ea4591e5e0dcfab27819,
it is possible to use `Ctrl+Space` to stop the on-going activity.
Leaving open, waiting for a single mouse-click way to achieve that too.

GeraldJansen:
Resolved by #427.

ederag:
Indeed, now it is also possible to stop an activity by a single mouse click, thanks @GeraldJansen !

PR #427, opened 2019-08-14
Add stop_tracking button to header bar

GeraldJansen:
This PR adds a stop_tracking button to the header bar of the overview window (#258). It complements the Ctrl-Space keyboard shortcut and does the same thing.

A few other tweaks are proposed: 1) the settings icon is replaced by an open-menu icon, 2) the order of the buttons is reversed to reflect current trends and to be more similar to apps such as yelp.

### DIFF
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -70,20 +70,26 @@ class HeaderBar(gtk.HeaderBar):
         self.range_pick = RangePick(stuff.hamster_today())
         self.pack_start(self.range_pick)
 
-        self.add_activity_button = gtk.Button()
-        self.add_activity_button.set_image(gtk.Image.new_from_icon_name("list-add-symbolic",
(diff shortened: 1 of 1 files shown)


GeraldJansen:
Screenshot_2019-08-14_16-12-17

ederag:
Thanks, perfect !

On openSUSE the default "open-menu-symbolic" 
`/usr/share/icons/hicolor/scalable/actions/open-menu-symbolic.svg`
still looks like a setting wheel,
but with other themes (e.g. adwaita), it does look like your menu icon. So all good.

This is a really nice addition, at the right place.
Just in case you would like to add more, we might consider reviving the `ActionRow`. 
To get a sight of it, uncomment this line:
https://github.com/projecthamster/hamster/blob/cbb508f8e0e995283edcdbb40f2aace9f20fa7f8/src/hamster/widgets/facttree.py#L286

Issue #305, opened 2016-03-04
Cannot add acitivity description form the "update activity" window


sliwowitz:
I use the fedora 23 package which got updated to hamster 2. I cannot add descriptions to activites using the "update activity" window:
- if I start a new activity from the shell extension everything is OK:
  - I enter "name@category, description" and it works
- if I "add earlier activity", description is ignored (I enter it, click save, and it disappears)
- if I modify an existing activity which had no description (the "update activity" window), newly entered description is ignored again

axelsimon:
I'd like to add some information to this.
If you add some tags, then anything after the #tags gets dropped. If you put it before the tags, it gets saved as the category (stuff@category my description, #tag1, #tag2).

Using the trick in issue #295 however, one is able to add a description after the tags, by entering it in the description box.

GeraldJansen:
Is this not solved by #403 which fixed #295? I.e. the description is taken from the separate description input box, which is now visible. It is not clear to me that it is desirable to allow description input from two areas of the "Update activity" and "Add earlier activity" widgets. In, the current behaviour is a bit strange for updating or cloning an activity with a description in that the description is replicated in two places, making it unclear where any edits should be performed. For example:
![Screenshot_2019-05-19_20-00-25](https://user-images.githubusercontent.com/31414956/57986111-fa8d4980-7a70-11e9-8b7d-46d4e3d330c1.png)

sliwowitz:
I have not used hamster for quite some time, as it, unfortunately, went downhill - first the usability with bugs like this one, and ultimately the shell extension not working on any of my machines (recent Ubuntus and Fedoras).

Anyways I just installed it to test it - the separate description input is cool, and updating the description works from there.

The top input which shows the already added description is strange indeed, and it seems to me that it's still not working - i.e. If i change or add the description in the top input field, it is still not saved.

ederag:
I have been  working on it, fully agree with both of you, and need feedback on the design.
Currently, I would tend to
1) Accept a description entry "inline" (after comma),
  in order to allow people used to the command line to get a familiar workflow
  (they will often be tempted to enter it this way, and it is not good to ignore an entry, IMO)
2) Only show the description in its box, once validated
  (e.g. when editing an existing activity)
3) If a description is given inline (after comma), with another one already in the box,
  if save is attempted, raise an error window and do nothing.
  The error window message could be
  "Attempted to enter a description inline (after the comma), while another description is already in the description box".
[...]

PR #430, opened 2019-08-20
Fix update activity description ambiguity

ederag:
Allow entering description 
either from the activity line or from the description box,
but not both (fix issue #305).
Prevent saving and give warning in case of ambiguity.

In activity line only, "valid" check mark icon to the right
Screenshot_20190820_121643

Both in activity line and in description box: ambiguity (which one to use ?),
"error" check mark icon
Screenshot_20190820_121902
More information in the icon tooltip (hovering mouse on the icon)
Screenshot_20190820_122124

In description box (multiple lines allowed there), 
"valid" check mark icon
Screenshot_20190820_122457

When editing an existing fact, the description is now only displayed in its box.

### DIFF
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -82,10 +82,12 @@ class CustomFactController(gobject.GObject):
                 original_fact = None
 
         if original_fact:
-            label = original_fact.serialized(prepend_date=False)
+            stripped_fact = original_fact.copy()
+            stripped_fact.description = None
+            label = stripped_fact.serialized(prepend_date=False)
             with self.activity.handler_block(self.activity.checker):
                 self.activity.set_text(label)
-                time_len = len(label) - len(original_fact.serialized_name())
+                time_len = len(label) - len(stripped_fact.serialized_name())
                 self.activity.select_region(0, time_len - 1)
             buf = gtk.TextBuffer()
             buf.set_text(original_fact.description or "")
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -59,6 +59,7 @@ class CustomFactController(gobject.GObject):
 
         self.description_box = self.get_widget('description')
         self.description_buffer = self.description_box.get_buffer()
+        self.description_buffer.connect("changed", self.on_description_changed)
 
         self.activity.grab_focus()
         if fact_id:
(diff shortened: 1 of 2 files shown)


ederag:
On second thoughts, the entry icon does not add much.
The information tooltip should be set on the save button,
that would be more discoverable ?
What do you think ?

ederag:
Done, much better, isn't it ?

GeraldJansen:
Looks good and works well!

My only nitpick is that the description should probably be excluded from all auto-completion suggestions, including the one with [stop now], when the fact discription box already has some content.

GeraldJansen:
Regarding [How to contribute](../wiki/How-to-contribute), this is what works for me while avoiding fetching tons of old pull requests:

Add this line to [remote "upstream"] section of .git/config:
`        fetch = +refs/pull/*/head:refs/remotes/upstream/pr/*`
Fetch and checkout PR 430, for example:
`git fetch upstream pull/430/head && git checkout pr/430`
....
After testing, delete your local pr/430 branch:
`git checkout master && git branch -D pr/430`

Note that I have origin as my fork and upstream as projecthamster/hamster and this is probably what most would have if they followed the forking instructions.

(I know this is not the right place to discuss this but the developers' mailing list is ... dead?)

ederag:
Good catch ! Done.

About the "how to contribute" instructions,
the best course of action might be to open an issue
(i receive notifications for anything posted on projecthamster/hamster).
Then I would add the label "needs doc" in this case, or "meta" for general discussions.

ederag:
Good catch also for the instructions, wiki [updated](https://github.com/projecthamster/hamster/wiki/How-to-contribute/_compare/035b5721acda4321a6930fe8599775975819aafc...477a11a47dba0a9a7dd92e97ea84c129803a59bf), thanks !

sliwowitz:
I like this idea, but for whatever reason, I'm not getting the warning icons - the Save button gets greyed out when I enter a duplicate description, but the entry line only shows the down arrow and not the checkmark/x (and I get no warning tooltip either). This is on Ubuntu 19.04.

Screenshot from 2019-08-23 12-21-42

ederag:
@sliwowitz Thanks for the feeback. 
Indeed, the icons have been removed, they were distracting [IMO](https://github.com/projecthamster/hamster/pull/430#issuecomment-522959653).
And a tooltip on the icon would have been much more difficult to find,
while attaching it to the save button, it will appear when trying to save.

sliwowitz:
Ah, you're right. I missed your earlier comment and went just by the pictures. I'd say it's good then, I've been playing with it for past couple hours, and I like it.

ederag:
Then thanks for posting the screenshot, that will hopefully prevent future readers from being puzzled.

Merging, thanks for the feedback !

PR #436, opened 2019-08-31
Clarify hamster cli usage notes

GeraldJansen:
This PR attempts to clarify the usage notes printed by `hamster[-cli] -h` which are currently somewhat misleading.

For example, the current help implies that one can specify the start and end times, and even though they would be parsed correctly, the program actually only uses the date part and not the HH:MM part. 

The message about missing start and end times was also misleading (00:00 to 23:59 instead of using preference for start of hamster day, eg. 05:00).

### DIFF
--- a/src/hamster-cli
+++ b/src/hamster-cli
@@ -360,10 +360,10 @@ if __name__ == '__main__':
 Actions:
     * start / track <activity> [start-time] [end-time]: Track an activity
     * stop: Stop tracking current activity.
-    * list [start-time] [end-time]: List activities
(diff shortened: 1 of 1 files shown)


ederag:
Indeed, much better, thanks !
Some work will be required to improve consistency (#438).
I'm working on an important refactoring of `Fact`,
then on these date issues that will depend on it.

Side note: this the end of my "holidays", expect much lower availability in the next months.

GeraldJansen:
Thanks for all that holiday work!

Issue #419, opened 2019-06-26
Tasks in the morning are storen on the wrong day


ego2dot0:
If start a task early in the morning (<5am), the task is added the day before, instead of the current day. Correcting this task is a problem too. I have to add I it tomorrow in the edit view, to insert it today.

This behaviour overrides all the early day tracking of the day before.

GeraldJansen:
This is a feature for people that work late at night and into the early hours of the next day. 

You can change the behaviour in the settings (Time Tracker Preferences -> New day starts at: 05:00).

ego2dot0:
That feature has an bug. If I add a task from 04:00 - 06:00 at 16th (15th day before) of June. It will be added at 14th of June. Same if i start a new task in the night. It will be placed 24h hours before and overrides existing times.

ederag:
Thanks for the screenshots.
On current master (0ddce419f4e367f27b76dff6eb6494945bbd7fe0),
the activity overlapping end-of-day time appears at the "correct" day,
i.e. the hamster day of the start time.
(that would be the 15th in your example, not the 14th as in the screenshot).
So previous data would not be overwritten.

What is your version ?

Nevertheless, in master there is a remaining bug because the duration is negative.
I'll look into that. Thanks again for the report !

pwhipp:
The negative duration will disappear if you restart the app

ederag:
> The negative duration will disappear if you restart the app

@pwhipp Can't reproduce (negative is persistent here). Which version are you using ?

pwhipp:
@ederag 
Hmm... no version in 'help about' :| It was master.zip downloaded on July 8th.

I did also set the day start time in the tracking settings appropriately before restarting (e.g. 02:30 in my case because I often have to start work at 04:30).

ederag:
@pwhipp So we should have the same results. 
Thanks for giving the day start time.
What were the start and end times of the activity ?

pwhipp:
@ederag I can't remember but it would have started with the default day start time and I regularly start work between 04:30 and 05:30 so I'm guessing the offending times were in that range.

ederag:
> I did also set the day start time in the tracking settings appropriately before restarting 
(e.g. 02:30 in my case)

@pwhipp Got it; after restart the activity was not overlapping the hamster day start time any longer.
Yet a change in the hamster day start time was taken into account only after hamster restart.
Actually just changing the date (left/right arrows) back and forth seems to be enough to update the display.
That would be a different bug.
[...]

PR #437, opened 2019-09-01
Fix dates

ederag:
This hopefully fixes date-related issues

In particular #419 is solved by using a clear rule rather than heuristics:
for activities overlapping the hamster day-start time, use
the selected hamster-day for start_time
(since it is an early morning start, this corresponds to the next civil date), 
and the next date for end_time.
A warning informs the user that something unusual is happening,
and gives full information to check:

image
The displayed duration is correct (not "negative" any longer).

If the fact is to be created in the past, the selected date has to be decremented.
It is much safer to create the next civil day by default, since chances for conflicts is reduced.

Note that now the overlapping fact is displayed for the selected hamster-days (26th here),
because part of its activity belongs to the 26th.
Its indicated date is the start-time hamster-day.

Bonus: on-going facts from the month before displayed dates are now shown without the need
for a second search.

### DIFF
--- a/src/hamster-cli
+++ b/src/hamster-cli
@@ -218,7 +218,7 @@ class HamsterClient(object):
 
         start_time = start_time or dt.datetime.combine(dt.date.today(), dt.time())
         end_time = end_time or start_time.replace(hour=23, minute=59, second=59)
-        facts = self.storage.get_facts(start_time, end_time, ongoing_days=31)
+        facts = self.storage.get_facts(start_time, end_time)
 
         writer = reports.simple(facts, start_time.date(), end_time.date(), export_format)
 
@@ -284,7 +284,7 @@ class HamsterClient(object):
 
     def _list(self, start_time, end_time, search=""):
         """Print a listing of activities"""
-        facts = self.storage.get_facts(start_time, end_time, search, ongoing_days=31)
+        facts = self.storage.get_facts(start_time, end_time, search)
 
 
         headers = {'activity': _("Activity"),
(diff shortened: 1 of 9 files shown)


ederag:
[Good catch](https://github.com/projecthamster/hamster/issues/419#issuecomment-527548830), and good guess @GeraldJansen, thanks !
Found the culprit. Answering there to remember to fix it properly before merging.

ederag:
Can't reproduce here. Cleaned up the potential culprit.
Could you please try 7a8931af90a5a28e49aa89f79480b49a3f94789f out,
from a terminal (to see the exception) and report back ?

By the way, the default start time is 5:30 in `configure.py` (that would be the "true" default value),
not 5:00 as implicit in the code removed from `db.py`.

GeraldJansen:
Confirm resolution of the problem. No exception was raised for use of conf.day_start.

However, activities started exactly at day_start (eg. 04:00) are still shown on two days. This little fix in db.Storage._get_facts resolves that:
`datetime_to = dt.datetime.combine(end_date, split_time) + dt.timedelta(days=1, seconds=-1)`
Before:
`hamster.storage.db - INFO - searching for facts from 2019-09-05 04:00:00 to 2019-09-06 04:00:00`
After:
`hamster.storage.db - INFO - searching for facts from 2019-09-05 04:00:00 to 2019-09-06 03:59:59`

Thanks for correction of default day_start.

ederag:
Indeed, thanks !
(forced push to fix a typo in the commit message)

Issue #423, opened 2019-07-17
existing activities starting with the # character are displayed blank and new activities starting with # cannot be added


pwhipp:
Try adding an activity and typing # as the first character. The Save option is enabled but clicking it has no effect.

If you import a db (or update one for testing) with the # character as the first character in an activity, the activity appears as a blank entry.

ederag:
Confirmed, thanks for the report !
Way to reproduce, from the command line:
```bash
> hamster track #123
Error: please specify activity
```

mwilck:
IMO this is not a bug. '#' as first character indicates a **tag**. Every person who as ever used or read twitter should be able to grasp that concept. Why is it important to be able to have activities start with '#'?

GeraldJansen:
Because #nnn is also often used to refer to an issue or PR in some issue tracking software I suppose.  In the past there has been work on integrating Hamster with Redmine, for example.

mwilck:
We can of course discuss exchanging '#' for a different character. We could even make it configurable I guess. But the choice of '#' has always made a lot of sense to me, personally.

Hey, myself I'm working on such issues almost all day. I use '123456@Issues'. I've never felt the need to use '#123456@Issues' instead. Sure, that's a matter of taste. But how long is it that hamster has been using '#' for marking tags? Probably as long as tags have been supported at all.

mwilck:
@GeraldJansen, has that Redmine integration work ever lead to anything usable? Is it an active project?

GeraldJansen:
@mwilck At least one group has used Redmine integration in the past. The project is inactive, like Hamster itself for lengthy periods ;-) Actually, I don't even know if they used the #nnn in activity names. I was just trying give some possible reasons for this request. I guess it is a "bug" if it worked in the past and no longer works as expected.

ederag:
#466 should allow for activities starting with a `#`.
Beware that doing so, you'll be confined to separate fields for this fact.
(losing quickness of the cmdline, cf. #467)
That is tracked in #465 and will be hopefully be solved in a future version.

ederag:
Should be fully fixed by the #475 parser change.

ederag:
Better fix in #482.

PR #440, opened 2019-09-08
Verbatim database reading

ederag:
Prior to this PR facts read from the database were serialized and parsed again.
With this PR facts are read verbatim from the database.
This is expected to partially solve issue #423,
and clarifies the code.

`activity`, `category` and `description` are now properties, with setters to sanitize them.
They are now guaranteed to be strings (empty meaning none).

Breaking change: 
`Fact(str)` has to be replaced by `Fact.parse(str)`

### DIFF
--- a/src/hamster-cli
+++ b/src/hamster-cli
@@ -319,7 +319,7 @@ class HamsterClient(object):
 
         by_cat = {}
         for fact in facts:
-            cat = fact.category or _("Uncategorized")
+            cat = fact.category or _("Unsorted")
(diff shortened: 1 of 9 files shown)


GeraldJansen:
I have tested this PR for the last two days and not run into any problems. I also used it in combination with my [htool](https://github.com/GeraldJansen/hamster-tool) to completely rebuild my database, trimming it to the last 3 years of data. I checked that I could regenerate identical billing reports using the rebuilt db and all seems fine.

PR #336, opened 2016-10-04
Added an appdata.xml file for Linux software gallery integration

Mailaender:
See https://people.freedesktop.org/~hughsient/appdata/ for more information.

PR #441, opened 2019-09-13
Add appstream metainfo.xml file.

GeraldJansen:
Added an metainfo.xml file for Linux software gallery integration. 

This is simply an update of PR #336.

Co-authored-by: Mathias Mailänder (@Mailaender).

WIP: I am not 100% sure this works. I tried to validate the xml with `appstreamcli validate` and got some warnings (eg. screenshots don't yet exist at their eventual destination in master). After installation (in /usr/share/appdata) and  running `sudo appstreamcli refresh --force`, I still didn't find hamster or "time tracker" in Ubuntu software. Tried moving the file to /usr/share/metainfo where most apps now put the xml files (on Ubuntu at least) but no joy.


GeraldJansen:
@Mailaender if you could send me your no-reply.github.com email I will put that in the commit message.

Mailaender:
Maybe mailaender@users.noreply.github.com

data/hamster-time-tracker.metainfo.xml
```diff
@@ -0,0 +1,53 @@
...
+  <provides>
+    <binary>hamster</binary>
+  </provides>
+  <releases>
+    <release version="2.2.2" date="2019-02-11">
+      <description>
+        <p>Python3 pre-3.6 compatibility</p>
+      </description>
+    </release>
+  </releases>
```
ederag: Are these lines useful ?
From the linked [documentation](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps):
> They are not strictly required to be present

They would increase the maintenance burden,
and a failure to update these fields could give the wrong impression that the installed version is obsolete.
There is probably a github API to create a release that we could use to ensure consistency,
but I'd vote to simplify our lives.
Unless this information is really useful there ?

GeraldJansen:
Thanks @ederag. All suggestions accepted.

ederag:
Thanks to both of you !
@Mailaender there is no hurry (after about three years !),
I'm just merging right away to get the screenshots online,
so that Gerald can update the locations.
Please let us know here or in the forthcoming PR 
if any of the changes were not appropriate.

Issue #64, opened 2013-01-13
Migrate from GConf to GSettings


eheintzmann:
With GNOME 3.0 (april 2011), the GNOME Project decided to discontinue GConf.
It won't be maintained anymore: bugs and security holes will not be fixed.
Thus, Hamster shouldn't depend on GConf any longer.
Please, migrate to GSettings.
GStettings is the official GNOME replacement for GConf, and the GNOME project
is porting all of its applications to it:
https://live.gnome.org/GnomeGoals/GSettingsMigration

You will find the official from GConf to GSettings porting guide at:
http://developer.gnome.org/gio/stable/ch31.html

ederag:
Here is an example of a migration to gsettings:
https://github.com/xflux-gui/fluxgui/pull/112/commits
especially
https://github.com/xflux-gui/fluxgui/pull/112/commits/aea7d3e32ebbdeec472f95fc8a344c8c9ea30ebe and
https://github.com/xflux-gui/fluxgui/pull/112/commits/4cb62e3acb2761318e6e4999f8b629251f5c973d
[there are other relevant commits for settings in this PR]

GeraldJansen:
Migrating from gconf to dconf should be done for v3.0 or v3.1 if they should become  "the stable version for years to come"(https://github.com/projecthamster/hamster/issues/433#issuecomment-524579649)

ederag:
No gsettings for for `v3.0`.
OK for `v4.0` (that is a major change).

City-busz:
This would be important on Arch Linux, because we are trying to get rid from GConf, and I had to remove Hamster from the official repositories because of this.

ederag:
Sorry.  This is a huge change, a new framework to me, and too many jobs already started.
I was hoping to get another year to tackle this task...

If anybody knows how to do that, please fork, and advertise your progress here,
so that Arch Linux users can readily benefit from your work.

Ideally, do this work in a separate branch (say `gsettings-migration`), 
with only the minimal changes required to migrate without introducing bugs.
This would help creating a clean PR (please be nice to reviewers !).

The configuration machinery is located in
[`src/hamster/lib/configuration.py`](https://github.com/projecthamster/hamster/blob/master/src/hamster/lib/configuration.py)
For the installation, waf has the [glib2](https://waf.io/apidocs/tools/glib2.html) tool.
[...]

PR #443, opened 2019-09-17
Remove dependency on gconf

GeraldJansen:
- completely remove gconf dependency (also in waf scripts)
- save/load configuration in xdg_config_home, normally ~/.config/hamster-time-tracker/hamster.json
- for now the file is accessed on every conf.set() and conf.get() call (meaning that external modifications will be picked up immediately)

Related to #64. Perhaps a complex migration is not necessary.


ederag:
Sorry, [as i said](https://github.com/projecthamster/hamster/issues/64#issuecomment-531505437), this will have to wait after v3.0.

GeraldJansen:
Okay, I guess it's ready when you are.

ederag:
OK, it was not clear that it was work in progress, 
and I do not want to give the impression that PR are ignored, especially PR from you :smile:.
I understand the value of this work, both as a proof of concept, 
and as a solution for distributions that are going to drop gconf.

But really I feel that handling configuration (with eg. file monitoring), is no small feat.
Gsettings migration is probably not that difficult, since its documentation about migration is a bit scarce,
and that would be a standard, well debugged solution.

What I would recommend is **not** implementing fancy file monitoring,
keep this proposition as simple as possible, to have a branch working around gconf disappearance.
But maybe that was your intention all along ?
If you go too far, I will feel compelled to use so much work done
(that's already the case, even with the warning given).

GeraldJansen:
My proposition is to move to very simple config file to handle our handful of configuration values. The file is re-read or re-written for every single access to any setting (i.e. conf.get('setting_name') or conf.set('setting_name, value), so no other file monitoring is needed. Not pretty or efficient or robust, but this little application doesn't need anything more complicated or robust, just like it doesn't need a more robust DB than sqlite3 for the real data storage. I have tested this, including external editing of settings, which are picked up immediately (well, the next time hamster needs to use the modified parameter).

ederag:
Even more interesting, but still need to think, later.

bgamari:
I'll admit, I'm not sure I see the rationale for this direction. It seems like the course of least resistance here would be to simply move to GSettings. Adapting the schema should be trivial and the GSettings interface is reasonably similar to GConf.

As has been pointed out, configuration management is not trivial and the ad hoc solution proposed here almost certainly won't be as robust as using a library. Let's avoid reinventing the wheel if possible.

GeraldJansen:
Okay, forget it. Release a new version of a Gnome software in 2019 still depending on gconf when Gnome itself abandoned gconf in 2011. Looking forward to someone else's better solution.

ederag:
Dear @GeraldJansen 
> Release a new version of a Gnome software in 2019 still depending on gconf when Gnome itself abandoned gconf in 2011

Looks like a direct criticism to me.
This is unfortunate, true.
But please understand that all distributions were shipping it until recently.

Maybe it should be added that I did look into this configuration stuff maybe a year ago,
and again before the waf update.
GSettings migration was not obvious, 
but others managed to do it, and it is standard (this is my point of agreement with @bgamari).
and while reading about that (in particular why they abandoned gconf),
I learned enough to be a bit scared of rolling our own one.

Your proposition is interesting too, 
it has some elegance (apart from the frequent disk accesses).
Please do not delete it.

On the contrary, I would really encourage you to merge master into this PR often,
so that others can have a working solution until the decision is taken.

There might be a source of misunderstanding:
all the work you see emerging now does not mean I have a lot of time,
nor that it has higher priority, 
but it was started long ago, it makes a whole 
(for instance https://github.com/projecthamster/hamster/pull/461#issuecomment-547386843)
and I need to see it finished.

Hoping you understand and we can continue united;
as I'm really drained and in no state to take any decision about this.

GeraldJansen:
@ederag Sorry for the sharp response in a moment of frustration. Looking forward to #470. Great to see a new contributor, really!

bgamari:
Thanks @GeraldJansen! Quite alright, I think we've all been there :)

ederag:
@GeraldJansen It's OK, because your impetus usually helps this project, 
yet this times it was on the "uncomfortable" side.

Now I would not like this project to lose two motivated contributors,
so we'll see if it can both make it into v3.0 *and* be released before the ubuntu feature freeze.
**No promises** at all, warning has been given [earlier](https://github.com/projecthamster/hamster/pull/443#issuecomment-532325878), and the following months will be very busy.

PR #445, opened 2019-09-19
Add Ubuntu build dependency

Forage:
The waf build will fail without it and the suggested package in the error message (`install "python-devel"`) is not helping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,15 @@ pgrep -af hamster
 
 ##### Debian-based
 
-ubuntu (tested in 19.04 and 18.04):
+###### Ubuntu (tested in 19.04 and 18.04)
+Run-time:
 ```bash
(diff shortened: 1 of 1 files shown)


ederag:
The first part of the message was more informative
```
The distutils module is unusable: install "python-devel"?
```
But thanks a lot for improving this part.
Let's keep it simple, main dependencies on a single line ?

Forage:
The message at least set me on the right track to find the solution. Fixing the doc is easier for everyone though.

Fine by me to stick to one line. I assumed only the runtime dependencies were listed on purpose, but I suppose those who need to install those manually intend to build the package anyway.

What is the reason gnome-doc-utils and yelp are kept apart though?

ederag:
> What is the reason gnome-doc-utils and yelp are kept apart though?

Because the main application works without them. Just the help would be missing.

Thanks for upgrading the doc, this will provide a much smoother experience indeed.

PR #453, opened 2019-09-21
.desktop file updates

ederag:
The previous `Name` field ("Time Tracker") was fine when it was *the* gnome time tracker.
Since it is not part of gnome any longer, 
it makes more sense to add "Hamster" to the name seen in the applications menu.

Also a proposition to change the comment (not sure of this one, open to discussion).
"Trackers" are often not welcome, how to convey the idea that this one is under user control ?
Although there *is* some humor in the shadiness of "Project Hamster - your time tracker"...


GeraldJansen:
The name change sounds good. Alternatives for the comment could be "Hamster is time tracking for individuals" or "Your personal time keeping tool". 

Of course all the current translation strings will no longer be used in the installed desktop files and users will see the English version. One possibility would be to change the origin strings to the new values in all the translation files and leave the old values of the translations. Not sure about that though.

ederag:
Good one, thanks !

GeraldJansen:
LGTM

Issue #447, opened 2019-09-19
What does this button do?


Forage:
In the main UI there are four buttons at the top right.
![Screenshot from 2019-09-19 15-05-45](https://user-images.githubusercontent.com/550915/65247949-a6af7180-daf1-11e9-8818-c2b036cfc01b.png)
The second one does nothing what so ever. What is it's purpose?
It would be good to give them all a tool-tip to know what they do and how/when they are supposed to work.

GeraldJansen:
Yes, tool-tips would be great, as would improved documentation.  The + button opens the "Add activity" dialog (as do ctl-+ and ctl-n). The second button is a small stop sign and means "Stop tracking" the currently ongoing activity (if any; also performed by ctl-space).

Forage:
In that case, if possible, it would be better that the button is only enabled when there's an actual activity ongoing.

ederag:
Solved by @GeraldJansen in PR #454.
Thanks !

PR #454, opened 2019-09-21
Add some tooltips for overview buttons

GeraldJansen:
Add tooltips for the buttons on the overview header bar to improve discoverability for first time users (cf. #447)

### DIFF
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -73,7 +73,7 @@ class HeaderBar(gtk.HeaderBar):
         self.system_button = gtk.MenuButton()
         self.system_button.set_image(gtk.Image.new_from_icon_name(
             "open-menu-symbolic", gtk.IconSize.MENU))
-        self.search_button.set_tooltip_markup("Menu")
+        self.system_button.set_tooltip_markup("Menu")
         self.pack_end(self.system_button)
 
         self.search_button = gtk.ToggleButton()
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -91,6 +91,7 @@ class HeaderBar(gtk.HeaderBar):
         self.add_activity_button = gtk.Button()
         self.add_activity_button.set_image(gtk.Image.new_from_icon_name(
             "list-add-symbolic", gtk.IconSize.MENU))
+        self.add_activity_button.set_tooltip_markup(_("Add activity (Ctrl-N)"))
         self.pack_end(self.add_activity_button)
 
 
@@ -451,7 +452,7 @@ class Overview(Controller):
         self.fact_tree = FactTree()
         self.fact_tree.connect("on-activate-row", self.on_row_activated)
         self.fact_tree.connect("on-delete-called", self.on_row_delete_called)
-        self.fact_tree.connect("selection-changed", self.on_selection_changed)
+
         window.add(self.fact_tree)
         main.pack_start(window, True, True, 1)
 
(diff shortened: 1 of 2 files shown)


GeraldJansen:
Okay, think this is ready now.

ederag:
Thanks, looks good !
A slight concern was that the `+` icon might mean either `Ctrl-N` or `Ctrl-R`, 
depending on the selection state.
This is the purpose of 357926a460efbb2b7ae46dfa58e8042470f75bec, please review.
It might be better to state the specific shortcut first (eg. `Ctrl-N or Ctrl-+`) instead of the reverse.
Please do make this change if you feel so.

GeraldJansen:
Having different tooltips for the `+` icon, depending on whether or not an activity is selected, as well as multiple options for key strokes, could become confusing. Perhaps a single tooltip like "Add new (Ctrl-N) or resume selected (Ctrl-R) activity"  would be better. More detail could be given on the Input help page (eg. alternative `Ctrl-+` key, which btw is less convenient on an English keyboard, requiring also the Shft key).  It's a matter of opinion I suppose, so whichever you choose will be fine with me.

ederag:
> Having different tooltips for the `+` icon, depending on whether or not an activity is selected, as well as multiple options for key strokes, could become confusing. Perhaps a single tooltip like "Add new (Ctrl-N) or resume selected (Ctrl-R) activity" would be better.

Interesting. 
Should the tooltip say what is going to happen if the item is activated now, 
or give a general information ?

gnome [guidelines](https://developer.gnome.org/hig-book/unstable/toolbars-labels-tooltips.html.en) did not tell much in this respect.

From [ux-planet](https://uxplanet.org/tooltips-in-ui-design-f63e117aa3d1)
> Contains static information
> Users don’t expect tooltips to change from one instance to the next, so they are unlikely to notice changes in dynamic content, such as status information.

Actually the same text is found [here](https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-tooltips-and-infotips), together with more extensive guidelines.

> "Add new (Ctrl-N) or resume selected (Ctrl-R) activity"

Yet this would be ambiguous: if an activity is selected, clicking the button will not add new.
it would be possible to say
```
Add activity (Ctrl-+)
If an activity is selected, clone (resume) this activity (equivalent to Ctrl-R)
otherwise create a new activity (equivalent to Ctrl-N)
```

The text has to be very clear to avoid this kind of [confusion](https://ux.stackexchange.com/questions/55324/should-a-tooltip-on-a-stateful-widget-describe-the-current-state-or-next-click-a?rq=1&newreg=531ee7d503dc4b36adfb0c62e095f79c#comment186653_55710).

Need to think. Right now saying what clicking the button will do seems adequate to me,
but you are right that a static tooltip is usually recommended.

Or maybe we should use a status bar instead ?

Need to think.

Renaming the "use-case" label to "design".

> Ctrl-+ key, which btw is less convenient on an English keyboard, requiring also the Shift key

`Right Ctrl - Numpad +` works too here, this is the one I'm using most often actually.

GeraldJansen:
:+1: for being thorough! My final feedback/suggestion: simply use "Add activity (Ctrl-+)" as tooltip and leave the documentation of the Ctrl-N and Ctrl-R alternatives to the Input help page.

ederag:
> simply use "Add activity (Ctrl-+)"

That would be good, and light.

Another option would be a mix of static and dynamic:
```
Add activity (Ctrl-+)
If an activity is selected, clone (resume) this activity (equivalent to Ctrl-R)
otherwise create a new activity (equivalent to Ctrl-N)
```
with an emphasis (bold face) on either "clone (resume)" or "new" depending on the selection state.

Perhaps a bit too much code for the benefit.
That's surely overthinking, but nice for the user ?

GeraldJansen:
I've had a look at tooltips for various applications I use and they are universally very short. They should be hints rather than full documentation.

ederag:
My proposition would be an [infotip](https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-tooltips-and-infotips) rather than a tooltip.
I agree that "Add activity (Ctrl-+)" would be more standard.
Yet the `+` button does two related but different things, that might deserve the infotip.

Let's take some time to let other users give their feelings on this subject.

ederag:
Well, given the enthusiasm around my proposition,
rolled back, and adopted your suggestion.
This has the huge benefit to keep code simpler. Thanks !
Waiting for your check before merge.

GeraldJansen:
Looks good to me.

ederag:
That was super quick, thanks !

PR #456, opened 2019-09-22
Rename input dialog to "Add activity"

GeraldJansen:
As discussed elsewhere, the "Add Earlier Activity" dialog should be renamed to "Add activity" because it can be used to enter earlier, current or even future activities. Use lowercase "activity" to be consitent with "Update activity".


ederag:
Sorry it took so long to get back to this.
Looking at it with fresh eyes, it is exactly what should be done:
"Add" is the adequate term suiting both new and clone.
Thanks a lot !

Issue #339, opened 2016-10-29
'Add Earlier Activity' combo box doesn't work under Wayland


hedayat:
In "Add earlier activity" window under wayland (Fedora 25), the combo box doesn't appear.

elbenfreund:
Thank you for your feedback, it is much appreciated.

Project hamster and its various sub-components is currently undergoing some major changes. We prepare the introduction of a rewritten codebase for most of the underlying functionality. A direct consequence of this is that it is unlikely that any open/new bugs within the current/old codebase will be fixed (unless someone steps up and offers to do so) as most resources currently available will be invested in making the rewrite prime time ready.

ederag:
Thanks for the report.
What "combo bow" is this ? The drop-down menu for suggestions ?
Which version ?
Is is still the case with current master ?

hedayat:
Yes, the drop-down menu for suggestions.
I'm using Fedora 2.0 version. I've not tried current master yet. If needed, I should try.

ederag:
Thanks for the precisions. The bug is most probably still there.
If you want to look into that,
the activity entry widget is added to the edit activity window there:
https://github.com/projecthamster/hamster/blob/2e52d217daca4cd3e034fa619e18f871038aa639/src/hamster/edit_activity.py#L50
And the drop-down that does not show up, is ironically called popup:
https://github.com/projecthamster/hamster/blob/2e52d217daca4cd3e034fa619e18f871038aa639/src/hamster/widgets/activityentry.py#L206

As usual, it is advised to discuss here prior to making extensive changes.
But any chance it was https://bugzilla.redhat.com/show_bug.cgi?id=1264355 ?
Maybe these are also relevant: https://wiki.gnome.org/Initiatives/Wayland/GTK%2B

Feedback from users with a more recent version of Gtk3/Wayland would be useful too.

hedayat:
Well, to make my report more complete:
The bug still exists in Fedora 29, with GTK 3.24.1. It should be pretty recent GTK/Wayland combination.

And No, it was not that bug about misplaced windows. And to make the report more precise, it seems that the window actually appears, but it immediately disappears as if some key is pressed to hide it without selecting any entry.

But maybe the 'Popups and grabs' section in GTK Wayland wiki is related to this problem.

ederag:
Thanks for the precisions.
There is some absolute positioning in
[...]

Issue #439, opened 2019-09-03
Cumbersome to edit time [regression]


CendioOssman:
It seems the 2.x code lost the separate input widgets for start and end time. This makes it rather annoying to modify time as the previous interface had quick suggestions for common intervals.

It looks like this was removed in 697a03e59312c72135788ba77ce09bd0ac0c6e0d. But I don't see any rationale. Did someone consider the feature pointless? Or was it removed because no one had ported the supporting code?

Would a PR be accepted returning the old behaviour?

CendioOssman:
How it looks in 1.04:

![2019-09-03-152142_3840x1200_scrot](https://user-images.githubusercontent.com/6640916/64176972-cb071f00-ce5e-11e9-999c-6406d9c7d13a.png)

GeraldJansen:
Just a guess, but I think the new approach of specifying times on the activity input line was done to harmonize input in the hamster edit windows and the gnome hamster-shell-extension. 

Personally I really like the new method, because it is very keyboard friendly. With the current master branch (i.e. later than v2.2.2), the start-end times are highlighted when you open the update window. A single keystroke moves you to the start or end and a space deletes the time(s). New times can be entered as 4 digits without the ':'.  Together with other keyboard shortcuts (#230, #387) and a window manager shortcut to open the hamster overview window, I now mainly use hamster through the keyboard.

CendioOssman:
Hmm... Alright. I guess I'll have to test the new code and see how it feels.

For me I saw two fairly different use cases here, that the old UI catered to:

 - Starting a new activity, where a single input field with some syntax shortcuts was/is fantastic
 - Editing/correcting an existing entry where it was more important to quickly adjust just a single detail without having to risk other information (which hopefully were already correct)

WBTMagnum:
I'm still irritated by the single input field. I really like it in the hamster-shell-extension, where I can quickly start and enter tasks. I also see the point on the CLI. For editing past and running tasks in the panel I would prefer separate input fields.

Just my 2 cents

ederag:
@CendioOssman and @WBTMagnum, 
thanks for explaining well why separate fields for *editing* are important,
all the more if you are using overview mainly for editing.
As @GeraldJansen nicely expressed, the command-line style entry has some advantages though.

One idea would be to separate "Add" and "Edit", not satisfying because cloning is in-between.
Maybe we could combine them ?
Got some ideas, need some time before discussing them.

[...]

Issue #451, opened 2019-09-20
"Add earlier activity" dialog usage is not obvious


Forage:
Using the "Add earlier activity" dialog of the latest sources is of no use at the moment, but I'm having a hard time determining if it's by design or simply buggy.

![Screenshot from 2019-09-20 11-16-25](https://user-images.githubusercontent.com/550915/65315338-327bd900-db98-11e9-989a-6c5166a4f372.png)

It's a dialog I use quite a lot and I expect it to allow me to quickly add an activity that started, well, earlier, by specifying it's name and the starting time.
There's this timeline at the top that does not allow me to do anything more than changing the day (with the left/right arrow icons).

Clicking on the downwards arrow gives me a list of past items, but displayed at the top left of the screen, separated from the dialog completely.

These dropdown items have the options "[start now]", "[start activity -n minutes ago]" and "[clear]" but it is not clear to me how this is supposed to work.

If it's 11:25 now, how am I supposed to quickly start an activity at 09:15?
Am I really supposed to calculate the number of minutes myself? 
And how do I actually provide the time?
What is the (intended) purpose of the timeline at the top?

Using Ubuntu 19.04 (GNOME Shell).

GeraldJansen:
This dialog should be named "Add activity" because it can be used to add past, present and even future activities. It is the key interface to create new activity items for those who cannot or prefer not to use external programs like the Gnome shell plugin. Basic information on data input is given in the "Input" help page. By default the new activity will be started at the current time, as this is expected to be the norm for someone actively tracking their time usage. To specify an earlier time, you can enter a specific time in front of the activity as say 09:15, 09.15 or simply 0915. If you don't remember the exact time but think you've been working already for say 20 minutes on the current activity, you can enter that as -20 instead of giving the wall clock time. You can also specify the end time for completed activities, separated from the start time by either a '-' or by a space. Concerning time input, you may find some of the other open issues to be of interest.

In the README it is openly admitted that some of the nicety of the old GUI got lost in the transition to gtk3, basically because no-one stepped up to re-implement it. The ability to click on the time-line to set times is one of those things. However, if you have already entered some activities, you will see the time blocks highlighted on the time-line. I guess this feature can help to avoid entering overlapping activies. 

(Aside: a kinder and less offensive title would be ""Add earlier activity" dialog usage is not obvious". Please consider that there is only a single maintainer of this project and he has basically given up his last two summer vacations and much free time in the evenings for more than a year now to get this show back on the road.)

[...]

Issue #438, opened 2019-09-01
date/time entry consistency


ederag:
1) When `date` and `time` can be given explicitly together, e.g. `2019-09-01 02:00`,
they should be local time (the same as on the desktop wall clock),
as in `hamster-cli`.
This is seems to make more sense when entering both start and end datetimes:
`2019-09-01 04:00 - 2019-09-01 07:00`

2) When a day is selected separately, as with the gui timeline arrows in `activity_entry`,
then it means "hamster day" (that by default starts at 05:00).
This way the activity will appear in exactly this selected day when displayed
by the overview, `hamster list`, `hamster search`, ...

3) When only `time` is entered, the default date should be `hamster_today`.
Indeed, entering `23.50` half an hour later refers most probably to the current hamster day.
(using today would be very inconvenient here, think of entering `23.50 - 00.20`...).

4) Facts overlapping the hamster day start time (05:00 by default) are tricky.
e.g. entering `04:00 - 07:00` on 2019-09-01 at 07:30 would result in a
`2019-09-02  04:00 - 2019-09-01 07:00`, hence with a negative duration.
This has been solved for the gui in PR #437 by incrementing the end date 
(more secure than decrementing the start date) in case of negative duration.
A warning and a means to check and correct entry was given.
For command line this is a bit trickier. 
Just error on negative durations and require user to give full datetime information ?

GeraldJansen:
It would be great if the same fact entry and parsing rules could be used for command line entry, the edit_activity window and the hamster-shell-extension:
`[yyyy-mm-dd]  [HH[:|.]MM]  [HH[:|.]MM] activity@category, description #tag1 #tag2 ...`
OR
`[yyyy-mm-dd]  [-nnn] [-nnn] activity@category, description #tag1 #tag2 ...`

In the first form, specifying end HH:MM lower than start HH:MM indicates a time on the next day. The maximum duration of an activity specified in this way would be limited to 23h59m.

ederag:
To be clear, my point 1 was to replace the current syntax with
```
[yyyy-mm-dd] [HH[:|.]MM] [yyyy-mm-dd] [HH[:|.]MM] activity@category, description #tag1 #tag2 ...
```
The change would be major since the first date would not be the hamster day of the activity any longer.
Both `[yyyy-mm-dd] [HH[:|.]MM]` are datetimes, in local time.
IMO, there is no way to handle safely activities overlapping the hamster day start with the current syntax.
[...]

PR #461, opened 2019-10-05
Separate edit fields

### LINKED ISSUE #280: Comma in activity moves to description
Hi,

How can I enter a comma in a title of activity? I am inserting activities via GUI and I always forget that comma stops the activity title and moves everything after comma to a description. That is annoying. Maybe there is a purpose for that in CLI version but in GUI it is confusing.

### LINKED ISSUE #301: No completion for category (regression)
The 2.x overview window seems to have lost the auto-completion for categories, which makes it very difficult to get them right. It doesn't work in the shell extension either so there's no workaround there.

ederag:
This PR is first step towards verbatim entry.
To fix #280 and #423 some changes will be required in db.py (working on it).

Meanwhile, using completion code/widgets that were still around,
this PR brings back separate field entries, 
as well as completion for time, activity, category and tags.
That should fix #301 and #439.

Screenshot_20191005_224618

Feedback welcome !

To do:
- [x] End time popup is too large
- [x] Fix empty start time issues
- [x] Find more standard icons, and/or check icons availability, provide fallback ? cf. https://github.com/projecthamster/hamster/pull/461#issuecomment-541577877
- [x] separate "go-down-symbolic" and clear icons

### DIFF
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -342,7 +342,7 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkEntry">
+                  <object class="GtkEntry" id="category">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                   </object>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -366,6 +366,38 @@
           </packing>
         </child>
         <child>
+          <object class="GtkBox" id="tags box">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="orientation">vertical</property>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="halign">start</property>
+                <property name="label" translatable="yes">tags</property>
+                <attributes>
+                  <attribute name="style" value="italic"/>
+                  <attribute name="weight" value="bold"/>
+                </attributes>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="fill">False</property>
+                <property name="position">0</property>
+              </packing>
+            </child>
+            <child>
+              <placeholder/>
+            </child>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">5</property>
+          </packing>
+        </child>
+        <child>
           <object class="GtkBox" id="buttons box">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
@@ -431,7 +463,7 @@
           <packing>
             <property name="expand">False</property>
             <property name="fill">True</property>
-            <property name="position">5</property>
+            <property name="position">6</property>
           </packing>
         </child>
       </object>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -181,10 +181,13 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkEntry">
+                  <object class="GtkBox" id="start time box">
                     <property name="visible">True</property>
-                    <property name="sensitive">False</property>
-                    <property name="can_focus">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="orientation">vertical</property>
+                    <child>
+                      <placeholder/>
+                    </child>
                   </object>
                   <packing>
                     <property name="expand">False</property>
@@ -236,10 +239,13 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkEntry">
+                  <object class="GtkBox" id="end time box">
                     <property name="visible">True</property>
-                    <property name="sensitive">False</property>
-                    <property name="can_focus">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="orientation">vertical</property>
+                    <child>
+                      <placeholder/>
+                    </child>
                   </object>
                   <packing>
                     <property name="expand">False</property>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -205,6 +205,7 @@
                   <object class="GtkExpander" id="start date expander">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
+                    <property name="label_fill">True</property>
                     <property name="resize_toplevel">True</property>
                     <child>
                       <object class="GtkCalendar" id="start date">
@@ -277,6 +278,7 @@
                   <object class="GtkExpander" id="end date expander">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
+                    <property name="label_fill">True</property>
                     <property name="resize_toplevel">True</property>
                     <child>
                       <object class="GtkCalendar" id="end date">
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -193,7 +193,7 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkCalendar">
+                  <object class="GtkCalendar" id="start date">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="year">2019</property>
@@ -248,7 +248,7 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkCalendar">
+                  <object class="GtkCalendar" id="end date">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="year">2019</property>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -302,7 +302,7 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkEntry">
+                  <object class="GtkEntry" id="activity">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                   </object>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -359,7 +359,8 @@
                   <object class="GtkEntry" id="category">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
-                    <property name="secondary_icon_stock">gtk-clear</property>
+                    <property name="primary_icon_name">edit-delete</property>
+                    <property name="secondary_icon_name">go-down-symbolic</property>
                     <property name="completion">category completion</property>
                   </object>
                   <packing>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -292,51 +292,9 @@
             <property name="visible">True</property>
             <property name="can_focus">True</property>
             <child>
-              <object class="GtkBox" id="activity box">
-                <property name="visible">True</property>
-                <property name="can_focus">False</property>
-                <property name="orientation">vertical</property>
-                <child>
-                  <object class="GtkLabel">
-                    <property name="visible">True</property>
-                    <property name="can_focus">False</property>
-                    <property name="halign">start</property>
-                    <property name="label" translatable="yes">activity</property>
-                    <attributes>
-                      <attribute name="style" value="italic"/>
-                      <attribute name="weight" value="bold"/>
-                    </attributes>
-                  </object>
-                  <packing>
-                    <property name="expand">False</property>
-                    <property name="fill">False</property>
-                    <property name="position">0</property>
-                  </packing>
-                </child>
-                <child>
-                  <object class="GtkEntry" id="activity">
-                    <property name="visible">True</property>
-                    <property name="can_focus">True</property>
-                    <property name="secondary_icon_stock">gtk-clear</property>
-                    <property name="completion">activity completion</property>
-                  </object>
-                  <packing>
-                    <property name="expand">True</property>
-                    <property name="fill">True</property>
-                    <property name="position">1</property>
-                  </packing>
-                </child>
-              </object>
-              <packing>
-                <property name="resize">True</property>
-                <property name="shrink">False</property>
-              </packing>
-            </child>
-            <child>
               <object class="GtkBox" id="category box">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
-                <property name="margin_left">4</property>
                 <property name="orientation">vertical</property>
                 <child>
                   <object class="GtkLabel">
@@ -359,8 +317,49 @@
                   <object class="GtkEntry" id="category">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
-                    <property name="primary_icon_name">edit-delete</property>
-                    <property name="secondary_icon_name">go-down-symbolic</property>
+                    <property name="primary_icon_name">edit-clear-all-symbolic</property>
+                    <property name="completion">activity completion</property>
+                  </object>
+                  <packing>
+                    <property name="expand">True</property>
+                    <property name="fill">True</property>
+                    <property name="position">1</property>
+                  </packing>
+                </child>
+              </object>
+              <packing>
+                <property name="resize">True</property>
+                <property name="shrink">False</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkBox" id="activity box">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="margin_left">4</property>
+                <property name="orientation">vertical</property>
+                <child>
+                  <object class="GtkLabel">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="halign">start</property>
+                    <property name="label" translatable="yes">activity</property>
+                    <attributes>
+                      <attribute name="style" value="italic"/>
+                      <attribute name="weight" value="bold"/>
+                    </attributes>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">False</property>
+                    <property name="position">0</property>
+                  </packing>
+                </child>
+                <child>
+                  <object class="GtkEntry" id="activity">
+                    <property name="visible">True</property>
+                    <property name="can_focus">True</property>
+                    <property name="primary_icon_name">edit-clear-all-symbolic</property>
                     <property name="completion">category completion</property>
                   </object>
                   <packing>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -2,6 +2,9 @@
 <!-- Generated with glade 3.20.4 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
+  <object class="GtkEntryCompletion" id="category completion">
+    <property name="minimum_key_length">0</property>
+  </object>
   <object class="GtkWindow" id="custom_fact_window">
     <property name="can_focus">False</property>
     <property name="events">GDK_POINTER_MOTION_MASK | GDK_POINTER_MOTION_HINT_MASK | GDK_BUTTON_PRESS_MASK | GDK_BUTTON_RELEASE_MASK</property>
@@ -351,6 +354,8 @@
                   <object class="GtkEntry" id="category">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
+                    <property name="secondary_icon_stock">gtk-clear</property>
+                    <property name="completion">category completion</property>
                   </object>
                   <packing>
                     <property name="expand">True</property>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -202,12 +202,26 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkCalendar" id="start date">
+                  <object class="GtkExpander" id="start date expander">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
-                    <property name="year">2019</property>
-                    <property name="month">8</property>
-                    <property name="day">15</property>
+                    <property name="resize_toplevel">True</property>
+                    <child>
+                      <object class="GtkCalendar" id="start date">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="year">2019</property>
+                        <property name="month">8</property>
+                        <property name="day">15</property>
+                      </object>
+                    </child>
+                    <child type="label">
+                      <object class="GtkLabel">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="label" translatable="yes">start date</property>
+                      </object>
+                    </child>
                   </object>
                   <packing>
                     <property name="expand">False</property>
@@ -260,12 +274,26 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkCalendar" id="end date">
+                  <object class="GtkExpander" id="end date expander">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
-                    <property name="year">2019</property>
-                    <property name="month">8</property>
-                    <property name="day">15</property>
+                    <property name="resize_toplevel">True</property>
+                    <child>
+                      <object class="GtkCalendar" id="end date">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="year">2019</property>
+                        <property name="month">8</property>
+                        <property name="day">15</property>
+                      </object>
+                    </child>
+                    <child type="label">
+                      <object class="GtkLabel">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="label" translatable="yes">end date</property>
+                      </object>
+                    </child>
                   </object>
                   <packing>
                     <property name="expand">False</property>
--- a/data/edit_activity.ui
+++ b/data/edit_activity.ui
@@ -2,6 +2,9 @@
 <!-- Generated with glade 3.20.4 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
+  <object class="GtkEntryCompletion" id="activity completion">
+    <property name="minimum_key_length">0</property>
+  </object>
   <object class="GtkEntryCompletion" id="category completion">
     <property name="minimum_key_length">0</property>
   </object>
@@ -314,6 +317,8 @@
                   <object class="GtkEntry" id="activity">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
+                    <property name="secondary_icon_stock">gtk-clear</property>
+                    <property name="completion">activity completion</property>
                   </object>
                   <packing>
                     <property name="expand">True</property>
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -51,7 +51,7 @@ class CustomFactController(gobject.GObject):
         self.activity_entry = self.get_widget('activity')
         self.category_entry = self.get_widget('category')
 
-        self.cmdline = widgets.ActivityEntry()
+        self.cmdline = widgets.CmdLineEntry()
         self.get_widget("command line box").add(self.cmdline)
         self.cmdline.connect("focus_in_event", self.on_cmdline_focus_in_event)
         self.cmdline.connect("focus_out_event", self.on_cmdline_focus_out_event)
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -197,6 +197,9 @@ class CustomFactController(gobject.GObject):
                 self.cmdline.select_region(0, time_len - 1)
 
     def update_fields(self):
+        self.start_time.set_time(self.fact.start_time)
+        self.end_time.set_time(self.fact.end_time)
+        self.end_time.set_start_time(self.fact.start_time)
         self.start_date.date = self.fact.start_time
         self.end_date.date = self.fact.end_time
         self.activity_entry.set_text(self.fact.activity)
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -92,7 +92,7 @@ class CustomFactController(gobject.GObject):
                 self.fact = base_fact.copy(start_time=hamster_now(),
                                            end_time=None)
             else:
-                self.fact = Fact()
+                self.fact = Fact(start_time=hamster_now())
 
         original_fact = self.fact
 
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -97,6 +97,7 @@ class CustomFactController(gobject.GObject):
         # not at init time when cmdline might not always be fully parsable.
         self.cmdline.connect("changed", self.on_cmdline_changed)
         self.activity_entry.connect("changed", self.on_activity_changed)
+        self.category_entry.connect("changed", self.on_category_changed)
         self.tags_entry.connect("changed", self.on_tags_changed)
 
         self._gui.connect_signals(self)
@@ -150,6 +151,11 @@ class CustomFactController(gobject.GObject):
     def on_activity_changed(self, widget):
         if not self.master_is_cmdline:
             self.fact.activity = self.activity_entry.get_text()
+            self.update_cmdline()
+
+    def on_category_changed(self, widget):
+        if not self.master_is_cmdline:
+            self.fact.category = self.category_entry.get_text()
             self.update_cmdline()
 
     def on_cmdline_changed(self, widget):
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -185,6 +185,7 @@ class CustomFactController(gobject.GObject):
             self.update_cmdline()
 
     def update_cmdline(self, select=None):
+        """Update the cmdline entry content."""
         stripped_fact = self.fact.copy(description=None)
         label = stripped_fact.serialized(prepend_date=False)
         with self.cmdline.handler_block(self.cmdline.checker):
@@ -194,6 +195,7 @@ class CustomFactController(gobject.GObject):
                 self.cmdline.select_region(0, time_len - 1)
 
     def update_fields(self):
+        """Update gui fields content."""
         self.start_time.set_time(self.fact.start_time)
         self.end_time.set_time(self.fact.end_time)
         self.end_time.set_start_time(self.fact.start_time)
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -48,6 +48,8 @@ class CustomFactController(gobject.GObject):
         # None if creating a new fact, instead of editing one
         self.fact_id = fact_id
 
+        self.activity_entry = self.get_widget('activity')
+
         self.cmdline = widgets.ActivityEntry()
         self.cmdline.connect("changed", self.on_cmdline_changed)
         self.get_widget("command line box").add(self.cmdline)
@@ -142,6 +144,8 @@ class CustomFactController(gobject.GObject):
 
     def on_cmdline_changed(self, combo):
         self.validate_fields()
+        self.fact = Fact.parse(self.cmdline.get_text())
+        self.activity_entry.set_text(self.fact.activity)
 
     def update_status(self, status, markup):
         """Set save button sensitivity and tooltip."""
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -189,11 +189,17 @@ class CustomFactController(gobject.GObject):
             self.update_cmdline()
 
     def on_end_date_changed(self, widget):
-        if self.fact.end_time and not self.master_is_cmdline:
-            time = self.fact.end_time.time()
-            self.fact.end_time = dt.datetime.combine(self.end_date.date, time)
-            self.validate_fields()
-            self.update_cmdline()
+        if not self.master_is_cmdline:
+            if self.fact.end_time:
+                time = self.fact.end_time.time()
+                self.fact.end_time = dt.datetime.combine(self.end_date.date, time)
+                self.validate_fields()
+                self.update_cmdline()
+            elif self.end_date.date:
+                # No end time means on-going, hence date would be meaningless.
+                # And a default end date may be provided when end time is set,
+                # so there should never be a date without time.
+                self.end_date.date = None
 
     def on_end_time_changed(self, widget):
         if not self.master_is_cmdline:
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -56,8 +56,6 @@ class CustomFactController(gobject.GObject):
         self.cmdline.connect("focus_in_event", self.on_cmdline_focus_in_event)
         self.cmdline.connect("focus_out_event", self.on_cmdline_focus_out_event)
 
-        self.day_start = conf.day_start
-
         self.dayline = widgets.DayLine()
         self._gui.get_object("day_preview").add(self.dayline)
 
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -79,7 +79,9 @@ class CustomFactController(gobject.GObject):
 
         self.save_button = self.get_widget("save_button")
 
+        # this will set self.master_is_cmdline
         self.cmdline.grab_focus()
+
         if fact_id:
             # editing
             self.fact = runtime.storage.get_fact(fact_id)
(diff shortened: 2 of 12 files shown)


GeraldJansen:
Wow!
Just some quick feedback and first impressions:

Likes:
- cmdline input like before at top for rapid keyboard input
- snappy/snazzy update mirroring between cmdline and separate input fields

Dislikes:
- different activity completion algorithms on cmdline and activity input field (i.e. no frequency/recency ranking in the latter)
- much screen real estate for two calendars that would rarely be used 

Requests:
- instead of two calendars, maybe start-date and end-date buttons in front of respective time input fields would be nicer (by default they would show the selected date and if clicked a drop-down calendar would be shown) 
- if a category is provided in the category (not Unsorted), the suggestions for autocompletion in the activity field should only show activities associated with the given category

Error messages:
```[src]$ Running in devel mode
  File "src/hamster/edit_activity.py", line 187, in on_description_changed
    if not self.master_is_cmdline:
AttributeError: 'CustomFactController' object has no attribute 'master_is_cmdline'

  File "src/hamster/widgets/timeinput.py", line 276, in _on_key_press_event
    if event.keyval == gtk.gdk.KEY_Up:
...
AttributeError: 'gi.repository.Gtk' object has no attribute 'gdk'
```

ederag:
Thanks for the feedback !
The second error message looks fixed with 2d18aea4edbf0c6b53c98c398d05da30c396ad2e.
The first one never happened here. Hopefully fixed with d57eb94a8f0c909433923b7ddf7ef6c2c9bfcada ?

Glad you liked the update mirroring !

About the calendars size, at one point we could use the widgets that are still in dates.py.
That's a lot of added code complexity; 
not sure it is worth it, since this window is kind of transient (only for edition).
We'll consider that later, but not before v3.0 please; I'm close to exhausted.
Besides, I find the full information of the calendar (especially the day of week)
very useful as a check when working around day-start-time (no need to click to see that).
Finally, until someone finds how to fix #339, it seems wise to stay simple here.

> different activity completion algorithms on cmdline and activity input field (i.e. no frequency/recency ranking in the latter)

Actually that's a deliberate design choice:  
- For the most used quick entry (cmdline), primarily keyboard access,
  show only full activities already entered in the past, sorted by recency.
- For the separate edit fields, primarily mouse access,
show only the full list given in preferences, sorted alphabetically, 
which makes it much easier to find the expected one.

Mixing both would mean complex code and unsatisfying trade-offs, IMO.

> if a category is provided in the category (not Unsorted), the suggestions for autocompletion in the activity field should only show activities associated with the given category

This was also my first tentative, but sometimes the workflow would be 
"click on activity", the good one not showing if in different category,
"erase category", "select category", "select activity".

The rationale for the current design was to
- allow to select both activity and category in one go,
- while still allowing category filtering 
  (eg. just type `@category` to see all activities in the `category` bin).

Side note: it would be desirable to have a `TreeStore`-like view, but there's a [gtk limitation](https://gitlab.gnome.org/GNOME/gtk/issues/1541).
Despite the closed status, the issue is still valid, only `ListStore` can be used in our case, IIUC.


Thanks for the feedback and the propositions !
This is only the first pass, feel free to advocate;
looks like the right time to have a thorough discussion on this design.

GeraldJansen:
The first error message happens every time I update or resume an activity with a description.

ederag:
> The first error message happens every time I update or resume an activity with a description.

Can't reproduce here, and this is very strange.  The grab_focus should have fixed that.
Could you please check (for instance with `gitk --all`) that you are indeed after d57eb94a8f0c909433923b7ddf7ef6c2c9bfcada ?

GeraldJansen:
Oops. Sorry, I had fetched the last two commits but not merged them. The message has disappeared.

ederag:
Good, thanks for the quick feedback !

GeraldJansen:
A small detail, but on my system the time input fields show a missing icon. Suggested change at src/hamster/widgets/timeinput.py line 63:
```diff
-        self.set_icon_from_icon_name(gtk.EntryIconPosition.SECONDARY, "gtk-clear")
+        self.set_icon_from_icon_name(gtk.EntryIconPosition.SECONDARY, "go-down-symbolic")
```

GeraldJansen:
>  ... while still allowing category filtering  (eg. just type `@category` ...)

:+1: works ok with filtering (otherwise I have far too many `activity@category` combinationsl) 

> looks like the right time to have a thorough discussion on this design.

I'm afraid the only way to get any feedback is to make a preliminary release (say 2.99 or 3.0-beta1). It doesn't look like anyone else checks out the PR commits.

ederag:
> A small detail, but on my system the time input fields show a missing icon.

Thanks for the observation, added to the todo list in the head post.
Unfortunately, "go-down-symbolic" does not convey the information
that the current selection is going to be cleared.
Is there a more standard icon for that ?
Or is there a way to ship fallback icons with hamster ?

ederag:
> works ok with filtering (otherwise I have far too many activity@category combinationsl)

Since your first feedback, I have been thinking of a three-way behavior:
1) first click on icon clears and displays the full list (same as current behavior)
2) second click on icon filters on the already entered category, if any
3) third click hides the popup

That would allow filtering by mouse only (currently text has to be typed).
**But** for the second click to work properly, the selected category has to be correct.
Sometimes it is already correct (e.g. when switching between two activities in the same project),
but otherwise, the workflow would be
1) select category (widget situated *after* activity in the layout)
2) click activity icon twice (situated *before* in the layout) to select the activity.

Category should be situated before Activity in the layout for this workflow,
but that would not match the activity@category display we are used to.

Thoughts ?

ederag:
> I'm afraid the only way to get any feedback is to make a preliminary release (say 2.99 or 3.0-beta1). It doesn't look like anyone else checks out the PR commits.

Got to think about this.
Let's hope others will find time to join here.
I'd like to finish a promising branch, not online yet.
Then while testing this branch, there are all your nice PRs to more deeply review,
and smaller [todos](https://github.com/projecthamster/hamster/projects/1#column-4030930).

Perhaps a pinned issue entitled "Feedback needed",
linking to the hotter issues/PRs would help ?

> Other minor feedback: perhaps the description field should move down to the bottom.

The idea was to allow quick tab traversal from cmline to description, followed by enter,
as the quick keyboard way to enter a fact with multiline description.

Why would you prefer description at the bottom ?

Forage:
Thanks for the mention in the other bug, I'm not tracking PRs by default.

First of all: thank you for the work on this dialog! I can't count how many times I messed things up because I forgot to change the date/day for a previous activity. Not to mention a sore finger for having to keep clicking the arrow to go back a few weeks for that matter ;)
Separate edit fields is exactly what is needed to adjust things quickly.

I'd have to agree with @GeraldJansen on the date picker space. Since the end date would not be used in most of the cases (ongoing activity or a finished one won't likely span multiple dates) and the start date would be a default (current) date that doesn't need changing in most of the cases these two date pickers would be better off as a pop-up when clicking the date input field.
Is that change essential for 3.0? Not if you ask me.

I do hope the time input fields provide a time picker instead of being only a free text input field.

Why the clear buttons and only for only categories and activities?

Please don't add button icons. That's so [pre-2009](https://www.osnews.com/story/21935/gnome-to-drop-icons-in-buttons-menus/) ;)

About the activity list part of the discussion:
Please don't give one button/icon different functions, you'd hide the functionality and people will get confused when they happen to click it a second time by accident.
I'd suggest putting the category before the activity input field, since you want to limit the activities based on the category and people tend to work from left to right.
Next when showing the list of activities you only show the activities of that category, with a special item at the bottom of the list with a label like "show all activities" to show the rest in the other categories as well.

GeraldJansen:
Concerning the icon in the time selection field,  there's no need to clear the field unless a different time is selected, so "go-down-symbolic" looks okay to me. 

Regarding category and activity fields, @Forage's suggestions sound good.

Concerning placement of the description field at the bottom: 1) it is the only field that expands vertically upon window resizing, 2) I suspect it is rarely used (by most users).  Not critical, just an idea.

ederag:
Thanks for the interesting discussion.

@GeraldJansen 
> Concerning the icon in the time selection field, there's no need to clear the field unless a different time is selected, so "go-down-symbolic" looks okay to me.

We do need a "clear" action for the end time, to make the fact on-going.
Or maybe that should be a separate icon. Probably.

> Concerning placement of the description field at the bottom

With quick Tab traversal, the description box becomes quite nice.
Actually, this is my preferred option when (rarely) entering a description.
There is another, perhaps better reason to keep it there:
the description box is common to cmdline and separate entries.

About category and activity, I'm about to be convinced.

@Forage 
> I'd suggest putting the category before the activity input field, since you want to limit the activities based on the category and people tend to work from left to right.
Next when showing the list of activities you only show the activities of that category, with a special item at the bottom of the list with a label like "show all activities" to show the rest in the other categories as well.

Good one indeed, and since Gerald likes it too, adjudicated, thanks !

> ... date picker space ...

I'm currently not convinced that screen real estate is worth the increased code complexity.
We'll see later.

> I do hope the time input fields provide a time picker instead of being only a free text input field.

Time can be selected by mouse in the dropdown (15 min resolution)

> Why the clear buttons and only for only categories and activities?

There are clear buttons for time entries as well.
Actually consistency was a reason to add one to activity and category.

> Please don't add button icons. That's so pre-2009 ;)

Thanks for the tip.
No icons to buttons, except for the bottom row (OK, Cancel, ...) where they are standard.

> About the activity list part of the discussion:
Please don't give one button/icon different functions, you'd hide the functionality and people will get confused when they happen to click it a second time by accident.

The icon had a single function in my proposition: 
cycle through completions (empty, from category, all).

Yet I'm now convinced that a "go-down-symbolic" as suggested by Gerald to toggle the dropdown,
and a separate icon to clear the entry is better, to make purposes clearer, indeed. Thanks !

Forage:
Icons and text for buttons is no longer standard, even for those buttons at the bottom. It's either an icon or text. Since it's standard for those at the bottom to have text no icon should be used.

Which desktop do you use that you still see a combination of the two?

GeraldJansen:
XFCE4 still uses both icon and text on the bottom buttons (I checked a few native apps). Iirc, @ederag uses KDE.

Forage:
Darn. GNOME no longer does this for a long time now and I tend to find much cleaner.
What do you currently see in the settings/preferences Hamster dialog for the "Close" button at the bottom? I have no icon. Does there happen to be some kind of magic already in place that excludes them in GNOME?

GeraldJansen:
I can't be of much help because I'm old-fashioned and normally use XFCE. It could be that some of the icons depend on the icon themes installed and used (eg. humanity-icon-theme, adwaita-icon-theme).

Forage:
So it would already help knowing if you see an icon on the close button of that dialog at the moment ;) If you do than all is fine (since I don't see one), if you don't than its not added by Hamster to begin with.

GeraldJansen:
Okay, so I now took the time to fire up gnome and can confirm: no icons on the bottom buttons on gnome (and I didn't change any default settings), yes icons on xfce4.

Forage:
Only an XFCE confirmation would have been sufficient, but great, thanks! So go ahead and add those icons to the buttons, I won't see them anyway ;-P

ederag:
Thanks for the feedback.
Entry icons are now separated, and hopefully standard ?
Sometimes the icon have to be clicked twice, not sure there is an easy fix.
Next step is to put category before activity.

GeraldJansen:
First of all let me say that I'm hesitant to make any comment other than "looks good" for fear of further delaying a release ;-)

It's a very minor consideration, but personally I find the two buttons per input field excessive. Another option would be to insert a blank line in the dropdown lists for end time and category (not needed for start time in my opinion). I'll leave up to those who strongly support the separate input fields to give their feedback. Here's a screenshot of the current version (on xfce)...
Screenshot_2019-10-21_17-00-33-400

ederag:
Agreed, it does not look very pretty with the `adwaita` icon theme.
I was using the `oxygen` icon theme:
Screenshot_20191021_183030

The idea was that a blank line in the dropdown is harder to discover than a delete button.
Adding an Action would be clearer, but they tend to appear at the end of the dropdown list,
rather than close to the entry, which is unfortunate IMO. 
And that also requires click/search/click.

The clear icon was much faster to implement than
adding a stop/on-going button to the right of end time.

Yet your screenshot is a strong incentive to reconsider !
Besides, with all the help you gave to this project, I'd really want you to enjoy a nicer interface.
Unless someone chimes in with another solution/idea, I'd remove all the clear icons,
and add a stop/on-going icon for end-time. There is already a clear action for category.

Forage:
I don't quite understand the need for a clear button. Removing those all together and have people empty the input fields manually gives a much clearer UI without loosing that much.
If you insist on keeping them than removing all colour would calm things down a bit.

ederag:
Well, `Gtk.EntryCompletion` does not like the `go-down-symbolic` icon idea,
it does not seem possible to click this icon another time to remove the popup.
And there is no access to this popup (it is private).
=> only `edit-clear-all-symbolic` colorless icons.
The popup can only be dismissed by clicking outside the entry.
And there is no way to show the popup without clearing first.
`GtkEntryCompletion` has a lot of limitations; some might be bugs.
To achieve a better interface, a whole custom entry popup has to be rewritten. Not now...

> have people empty the input fields manually gives a much clearer UI without loosing that much

The goal is to be able to use the separate fields with mouse only.
Actually, for consistency with activity/category entries, 
I would favor removing the `go-down-symbolic` icons in `TimeInput`,
and attribute their functionality to the `edit-clear-all-symbolic` icon.

Why not use `go-down-symbolic` everywhere ?
The expected meaning of `go-down-symbolic` is to show potential suggestions,
for investigation before choosing one of them or keeping the current one.
Clear is destructive.
The proposed icons make that obvious.
What is not obvious is that clearing yields the dropdown,
but as I said, there were many dead ends (not all shared here), so I'm giving up on that.

GeraldJansen:
Looks acceptable to me. Great work on the category/activity coordination, thanks.

ederag:
> But nothing happens when I enter a new activity in the main field and press `@`. 

Sorry, that is not expected in this proposition.
These fields are "verbatim" entries, without any parsing done.
The "Show all activities" display `activity@category`, 
but `category` and `activity` are stored separately, and once selected,
are dispatched to the corresponding entry.

Actually once the verbatim branch (ready, but not pushed yet) is merged,
entering (**not** selecting in the "show all" list) `activity@category` will result in
an activity named "activity@category"...
This will consistently solve all issues calling for verbatim inputs, such as #423 or #280.
`@` has not been requested yet, but it might be only a matter of time...
The advantages and simplicity of this rule will hopefully compensate
the missing functionality you were looking for ?

To achieve what you want in this proposition, the workflow would be
1) type `activity` in the activity field,
2) `Shift+Tab` to get to category, and 
3) choose in the completion list.

Are you often entering a new category ? (always interesting to know of other use cases)
Would it be worth the increased code complexity ?

mwilck:
Actually once the verbatim branch (ready, but not pushed yet) is merged,
    entering (not selecting in the "show all" list) activity@category will result in
    an activity named "activity@category"...

@ederag, please don't! That's a terrible step back in usability!

mwilck:
Sorry for jumping on this late. TL;DR: Sorry to say it, but I'm not fond of this proposal. Rationale below.

Disclaimer: I've only been looking at the screen shot, apologies if that made me get something wrong.

### Don't show two text input fields for the same thing

The "activity", "category" and "tag" input fields are redundant and should be removed. Hamster users have used a single text input field for this for ages, I see no reason to change that. The autocompletion OTOH is extremely useful and should be kept. **If** we need a second input method for these items, it should be graphical one (such as a tree view facilitating selecting category/activity with the mouse, and a drop-down list with autocompletion for the tags). Having two text inputs for the same thing is not good UI design, IMO.

I suppose the rationale for the additional fields is improving usablity for new users. *However*, new users won't necessarily understand hamster's concepts of "activity", "category", and "tag", so I don't think it'll be very helpful.

My suggestion: keep the current free-form text field, and display a label with a suitable example below it (possibly with a button to remove the label). Or simply use a tool tip. Or a "first time" dialog explaining hamster's concepts in just a few words. Or any combination of those.

Lastly, if you really, really feel you must keep the separate input fields, place them directly below the free text field rather than at the bottom.

### Show a graphical input for time rather than date

Also, rather than eating up much screen space with the seldom-used calendar widget, I'd display a graphical control for the frequently used **time input**. Not sure about you guys, but what I'd need most is something like "shift start time by -10 minutes", because I frequently forget to update my activity in hamster when I start working something new, and need to fix that later. Such a control I'd probably use 20x a day, whereas the calendar widget - maybe twice a month.

Wrt the calendar / date input, I'm not against having a graphical input for that in general, but I'd show a button that would open a separate calendar dialog. That way the rarely used calendar wouldn't take up half of the screen space of the dialog (IMO the current proposal looks more like planning a vacation than entering daily activities).

GeraldJansen:
@mwilk W.r.t. your comments above and [here](https://github.com/projecthamster/hamster/issues/433#issuecomment-546268889), perhaps a simple solution would be to add the possibility to Expand/Collapse the edit_activity window. In collapsed mode there would be the single input "cmdline" (even without the separate "description" field). By default "Add activity" could start collapsed and "Edit activity" could start expanded. 

In the short term, I feel that it is important that there soon be an new release, in time for packagers to get it into the next Ubuntu and Fedora releases. Hopefully an acceptable solution can be found rapidly at this point. 

W.r.t. comma and # in the activity name, it would be nice if uniform and well documented "raw fact" parsing could be applied for input from the GUI "cmdline" and from the CLI. (If the hamster-shell-extension can also comply, so much the better.) See also discussion [here](https://github.com/projecthamster/hamster/issues/438#issuecomment-529860808).

mwilck:
@GeraldJansen, I agree. Not sure what the issue with Fedora and Ubuntu is, openSUSE still ships hamster 2.2.2, and the only really annoying issues I'm having are related to the GNOME extension.

mwilck:
Btw, one issue that I've had with the "edit/add activity" dialog is the timeline at the top. It looks nice, but it evokes the impression that it's interactive, that it would possible e.g. to mark a time interval by dragging the mouse in it, and have this time interval be used for the activity being edited. IMO that would be most useful if it worked, with **zero** additional screen space required. But I guess it would be a lot harder to get right than a few input fields.

GeraldJansen:
@mwilck Selecting start and end times from the timeline is #313 and is open but "future". It was also discussed briefly under #451.  (Personally I found it never worked with satisfactory precision in old Hamster anyway, but I guess some people liked it.)

ederag:
> Expand/Collapse the edit_activity window. In collapsed mode there would be the single input "cmdline" (even without the separate "description" field). By default "Add activity" could start collapsed and "Edit activity" could start expanded.

Good idea.
But people wanting to enter fancy activities starting with `#` or containing `,` would get frustrated
until #465 is done.
Or we manage to keep a memory of the revealer state between executions.

GeraldJansen:
> keep a memory of the revealer state between executions.

Easily done after PR #443 :-)

ederag:
@mwilck 

> I've only been looking at the screen shot, apologies if that made me get something wrong.

Indeed, let's hope that's the main issue here.
Trying it might make things clearer. 
But thanks for the feedback, interesting nonetheless.

> Hamster users have used a single text input field for this for ages, I see no reason to change that. 

Me neither, that's one of hamster's strengths.

> If we need a second input method for these items, it should be graphical one

Graphical inputs do not have the required precision.

> such as a tree view facilitating selecting category/activity with the mouse

Tried that, [dead end](https://gitlab.gnome.org/GNOME/gtk/issues/1541) on the Gtk side IMO.

> drop-down list with autocompletion for the tags

already back.

> Having two text inputs for the same thing is not good UI design, IMO.

Unless they have orthogonal purposes, which is the case here.

> I suppose the rationale for the additional fields is improving usablity for new users. 

Main purpose is orthogonality actually.
Yet having the separate fields to check what is entered on the cmdline is nice.

> place [input fields] directly below the free text field rather than at the bottom

The fields are in the same order as in the cmdline.
Except for the description that must be just after cmdline for tab traversal,
to allow for quick keyboard only input of multiline descriptions.
And it belongs to both types of inputs, so it looks natural there.

> eating up much screen space with the seldom-used calendar widget

Well you are not the first one bothered with screen space.
That will not change in the close future: too much code complexity for the reward IMO.
Is it really annoying, or does it just look superfluous ?

> "shift start time by -10 minutes"

The plan is to allow mouse wheel change of times. Later...

And remember that the default entry is still cmdline, 
where start time can be changed by typing `-10` to start 10 minutes earlier.

mwilck:
> >  If we need a second input method for these items, it should be graphical one

> Graphical inputs do not have the required precision.

I assume you've used the Google Calendar, or any similar app? I think it can be done. And even if the precision was limited to, say, 15minutes, it would still be useful. Minute-scale corrections could be done in the text field. But this is hard to get right, and a long term project. No doubt about it.

>  >    such as a tree view facilitating selecting category/activity with the mouse

> Tried that, dead end on the Gtk side IMO.

Sorry to hear it. A tree view without autocompletion might still be useful, but that's not essential.

> >    Having two text inputs for the same thing is not good UI design, IMO.

> Unless they have orthogonal purposes, which is the case here.

I disagree. I've made my argument before. I'm not going to fight about it, but I don't like it. What exactly is "orthogonal" about these? AFAICS it's 2 different ways to enter the same information.

> The fields are in the same order as in the cmdline.

Fair enough, but optically the text fields are separated too far, mainly by the huge calendar widget.

> Except for the description that must be just after cmdline for tab traversal,
> to allow for quick keyboard only input of multiline descriptions.

Which shows nicely that the concept of this new dialog is inconsistent. If you want to design it for "quick keyboard-only input" (highly appreciated by me), remove everything below the description, and add a "detailed input" button that would open an extra dialog with the other fields. Actually I like that idea a lot. It would also eliminate my "2 inputs for the same thing" concern (the "detailed input" dialog would not have the aggregate text field, only the individual ones).

> Well you are not the first one bothered with screen space.

I'm not so much concerned with screen space as such. Only with the fact that the almost useless calendar widget takes half of the area of the dialog. It's the wrong emphasis, IMO.

ederag:
> What exactly is "orthogonal" about these? AFAICS it's 2 different ways to enter the same information.

Indeed, same information. But their use cases are orthogonal.
cmdline:
1) quick keyboard input in one go
2) completions from past activities
3) interpreted (parsed)

fields:
1) mainly mouse control
2) completions from lists that can be modified in preferences
3) isolated control, what you see is what will get into the database.

> optically the text fields are separated too far, mainly by the huge calendar widget.

This is hopefully partly alleviated by the fields being always at the same positions.

> Which shows nicely that the concept of this new dialog is inconsistent. If you want to design it for "quick keyboard-only input" (highly appreciated by me), remove everything below the description, and add a "detailed input" button that would open an extra dialog with the other fields.

That would add several operations for the user: 
find the "detailed input button", click on it, find the window on the screen, find the control, ...
Not nice at all. The v2 interface is much better in this respect (less clicks).
This PR is about bringing back v1 functionality while keeping v2 quickness.

Finally, cmdline and fields are synchronized 
(a change in cmdline is instantaneously reflected on fields and vice-versa).
So there is no conflict.
Moreover, this smooths the cmdline learning curve for new users,
as they will be able to experiment and immediately see the results.

> Only with the fact that the almost useless calendar widget takes half of the area of the dialog.

The calendar is not "almost useless" to people with activities crossing the hamster day start-time,
in which case it becomes quite important, as a check.
Granted, the display itself could be reduced to a single line,
but mixing that with a calendar popup (functionality present in dates.py) does not work on wayland.
So this is postponed until gtk/wayland related issues (#339) are resolved.

ederag:
Another important argument: there is not only `new` and `edit`, but also `clone`.
And `clone` may be closer to `edit` or `new` depending on the amount of changes to make.
This would complicate any attempt to make cmdline/fields visibility automatic.
Even making the choice manual, and reusing the previous choice by default 
would not be nice at all: too often would the last choice not be the preferred one.

ederag:
@CendioOssman [Was [this](https://github.com/projecthamster/hamster/pull/461#issuecomment-546088283) an accurate answer to your comment ?]
@mwilck I hope the above answers were convincing.
This PR is far from perfect, but unless I missed a point,
it could be merged on Tuesday, IMO;
raised issues such as hiding the calendar will be considered later.

CendioOssman:
I'm afraid I agree with @mwilck, removing the ability to do "activity@category" would be a massive step backwords in usability. It sounds like you're suggesting it will be made to behave exactly the same as the new "Activity" field?

My use case is that I have a fixed set of categories, but very often new activities. Every entry has both activity and category specified.

So that means I often enter a new activity, meaning I don't get any help from the general auto-complete. But once I hit "@" I will definitely want to use an existing category. So here it is very important that I have auto-complete, both for efficiency and to avoid mistakes.

1.04 had perfect behaviour for this and I'd very much like to see that kept in some form in the new code.

mwilck:
> That would add several operations for the user:
> find the "detailed input button", click on it, find the window on the screen, find the control, ...

Done right, it would be exactly one extra mouse click (the button). The details dialog would be focused/raised automatically, so that no searching would be necessary. Given that the separate inputs require between 4 and 8 mouse clicks (or TAB keystrokes) to navigate between the input fields anyway, that doesn't look like a big deterioration to me.

Maybe better yet: Keep these parts of the dialog hidden by default, and simply display [expander controls](https://developer.gnome.org/gtk3/stable/GtkExpander.html) for them. So you  don't even need an extra window.

> Moreover, this smooths the cmdline learning curve for new users,

Then, please, offer a configuration option to hide these inputs for users that have climbed the learning curve (which would take most users no more than a few days, I suppose). If you use expanders, it'd be as simple as toggling their default state.

> The calendar is not "almost useless" to people with activities crossing the hamster day start-time,

That includes myself, I do this from time to time. The actual argument is in my comment [from friday](https://github.com/projecthamster/hamster/pull/461#issuecomment-546246755). Activities spanning several work days are rare, and thus the date input is used *much less* than the other inputs. Some users will never need it. Yet it takes so much space and awkwardly sits *between* the related "command line" and "separate" input fields.

**Please**, rethink. At least keep the calendar hidden by default, or move it to the bottom.

ederag:
@CendioOssman 
> removing the ability to do "activity@category" would be a massive step backwords in usability. It sounds like you're suggesting it will be made to behave exactly the same as the new "Activity" field?

You mean in the cmdline, right ?
In the cmdline, "activity@category" allows to filter on past activities.

> My use case is that I have a fixed set of categories, but very often new activities. 

Thanks for giving more details on your use case. 
Having more new activities than new categories seems natural, you are not alone at all.

With this PR, to enter a new activity, you would use the separate fields,
first category (with completion => no mistake) then TAB to activity and enter the new one.

I do understand that you would like to do that in the cmdline.
That's no small feat.
This PR will allow to simplify the cmdline completion a lot (not started yet).
Maybe then will other completions be brought back.
But that's a pandora box. You are asking for category, others will ask for activity and tags.

Question is, how often do you enter a brand new activity, with respect to reuse an existing activity ?
(what ratio ?)

ederag:
@mwilck 

> Given that the separate inputs require between 4 and 8 mouse clicks (or TAB keystrokes) to navigate between the input fields anyway, that doesn't look like a big deterioration to me.

Most of the times these separate fields will be used for one or two changes.     
For more extensive changes the cmdline is easier.
So this is not a demonstrative comparison.

> Done right, it would be exactly one extra mouse click (the button). The details dialog would be focused/raised automatically, so that no searching would be necessary. 

We strongly disagree here.
Locating the entry on the screen 
(which is impossible to do before the secondary window or the expander are shown)
takes a non-negligible attention IMO.

> Keep these parts of the dialog hidden by default, and simply display expander controls for them.

Expanders need precise clicking, not as user friendly as having the whole entries available.
The following argument also applies here.

> Then, please, offer a configuration option to hide these inputs

The arguments given in a [previous comment](https://github.com/projecthamster/hamster/pull/461#issuecomment-546638835) show that this choice will not always be good.
That would increase [friction](https://blog.learningbyshipping.com/2015/03/16/frictionless-design-choices/) a lot. 
There is also a warning about too much configuration in this link, that applies here IMO.

> **Please**, rethink. At least keep the calendar hidden by default, or move it to the bottom.

No need to stress so much on "please".
In general, calm, rational arguments, careful reading of my answers, 
and clear explanations of use cases achieve more with me.

I already said that the calendar stuff will be reconsidered later. Not now.

Have you tried the interface now ?
Is it that bad in practice ?

mwilck:
@ederag,

> Expanders need precise clicking, not as user friendly as having the whole entries available.

This is only true if all entries are necessary, or equally important. Which is IMO not the case, that's my point. Displaying controls that the user doesn't need is confusing and distracting, and likely causes more friction than one extra click.

About "need precise clicking" - I really can't take that argument seriously. GtkExpander widgets can be set to occupy the whole width of the parent widget, making them easier to click than most other widgets in the dialog.

> No need to stress so much on "please".

I was trying to be polite. And I was also trying to be rational. I read your answers, too. I just happen not to be convinced, and was trying to point out why. Apparently you got a different impression. I'm sorry about that. *I really appreciate your work for hamster*, no mistakes.

> Have you tried the interface now ?
> Is it that bad in practice ?

I'm not saying it's "bad". I say it's too much, too big, distracting.

Here's one benefit I see: the different autocompletion in the "activity" field, with it's "filtered by category" option. That's a nice complement to the autocompletion of the free text field, which only shows recent activities. I'm not sure whether the "autocompletion mode" of the free-entry field couldn't be likewise improved, though.

I find the behavior of the calendar widget irritating. My expectation was that clicking on a different date than today would add a date in the "cmdline" field. That's not the case - clicking on the date only changes the "dayline" widget at the top. I can see that the reason for this is that date setting / parsing from the cmdline in the "edit/add activity" widget never worked as it should. Yet with an explicit date widget, I expected this to be fixed. In general, it's a bit confusing that some of the  controls are linked with the command line field, while others are not.

Actually, call me dumb, but the dialog gave me no explicit clue what effect the date widgets might have. Yet when I clicked "save", the date was actually taken into account for the created/modified entry. I guessed so, but it wasn't at all clear to me as a "novice" user of this dialog. Admitted, it's no worse than in v2.2, but it isn't a lot better either (and unlike 2.2, the dialog now kind of encourages me to play with the date widgets).

I believe that this is a bit dangerous. It's easy to accidentally click on the calendar somewhere without noticing (did I mention it occupies a lot of space?), and then wonder why the item isn't showing in the overview (because it ended up being saved for a different day).

I hope you find this rational and clear. I'm trying my best.

ederag:
@mwilck Thanks for the feedback.

> Displaying controls that the user doesn't need is confusing and distracting, and likely causes more friction than one extra click.

Fully agreed in general. I did read this nice [article about the number of clicks](http://uxmyths.com/post/654026581/myth-all-pages-should-be-accessible-in-3-clicks) earlier.
In the present case though, after using it for a few weeks,
when I enter from keyboard in the cmdline, I hardly even look at the other entries.
Of course it might be a developer bias, 
and please believe that I'm not taking your argument lightly. This is indeed worrying.

> GtkExpander widgets can be set to occupy the whole width of the parent widget, making them easier to click than most other widgets in the dialog.

Interesting, I had another type in mind, with a small arrow to reveal/hide. Maybe not a Gtk one.

> I read your answers, too. I just happen not to be convinced, and was trying to point out why.

Thanks for that, well explained feedback is precious.

> That's a nice complement to the autocompletion of the free text field, which only shows recent activities. 

Good, thanks for confirmation.

>I'm not sure whether the "autocompletion mode" of the free-entry field couldn't be likewise improved, though.

I [already said](https://github.com/projecthamster/hamster/pull/461#issuecomment-546880690) that opening this pandora box should be reconsidered later.
Beware that wanting everything in v3.0 is a good way to never see that release.

> I find the behavior of the calendar widget irritating. My expectation was that clicking on a different date than today would add a date in the "cmdline" field. That's not the case - clicking on the date only changes the "dayline" widget at the top. I can see that the reason for this is that date setting / parsing from the cmdline in the "edit/add activity" widget never worked as it should. Yet with an explicit date widget, I expected this to be fixed.

Indeed, that's related to #438.
I need a break. Has it really to be solved for v3.0 ?

The initial plan was to show date in cmdline only if it did not match the day set in timeline.
(that is, only for facts overlapping day-start)
Otherwise the cmdline will be cluttered with the start and end dates.

> Actually, call me dumb, but the dialog gave me no explicit clue what effect the date widgets might have. Yet when I clicked "save", the date was actually taken into account for the created/modified entry. I guessed so, but it wasn't at all clear to me as a "novice" user of this dialog. Admitted, it's no worse than in v2.2, but it isn't a lot better either (and unlike 2.2, the dialog now kind of encourages me to play with the date widgets).

Could it be partly because you decided that the calendars were useless,
and did not want to see them here in the first place ?
They are "what you see is what will get into the database".

> I believe that this is a bit dangerous. It's easy to accidentally click on the calendar somewhere without noticing (did I mention it occupies a lot of space?), and then wonder why the item isn't showing in the overview (because it ended up being saved for a different day).

That accidental click is another valid argument.
It is true that changing dates is dangerous,
and that it is more prone to happen with calendars than with the timeline arrows.
The best fix for this danger would be to highlight the potential overlapping facts 
in the timeline. Or at least to display a warning on the save button.
That's also in the long term plans. 
Finding the overlapping facts has to be separated from the `__solve_overlaps` first.

Need to rest and think.

> I hope you find this rational and clear. I'm trying my best.

Both rational and clear. Overall, very useful feedback, thanks !

mwilck:
@ederag, I fully understand that you need a break and that you want to get this off the table.
So please let my nitpicking here not stop you. I feel we agree on most issues.

Yet there's one thing that occured to me about how this newly designed dialog might influence user expectations, that I'd like to ask you to think about.

In theory, the separate entry fields would allow to overcome certain restrictions on input format, such as #280. **And this we should not do**.

To elaborate: The current dialog allows entering "x,y" in the activity field. If "saved" is clicked, this is transformed into an actitivity "x" with description "y". _This will irritate users_. Yet, it is better than forcing the "x,y" activity name to be set in the hamster data base, because **that would break the equivalence between the cmdline and multi-field input**. An activity set this way couldn't ever be edited or re-entered using the cmdline (or worse, trying to do that would probably generate a different, new entry).

Thinking about it, more than any nitpicking about calendar widget layout, this is my main concern - the new dialog would encourage users to write a bug report that entering "x,y" in the activity field doesn't do what they expect, and such a bug might be quickly fixed without much thinking, breaking the command line on the way. _This, I fear, opens up a Pandora's Box of user expectations_.

For me, _being able to fully control hamster with the single command line is absolutely a key feature of hamster which we shouldn't even think about changing._

Therefore I suggest that you minimal syntax checking be added to the activity and category fields, and indicate an error (e.g. by changing the bg color to red) when e.g. a comma is entered in the activity field.

I wouldn't consider even that a show stopper, but IMO it should be added quickly after a 3.0 release, to avoid user expectations to grow into the wrong direction.

mwilck:
Here's one more nit - try entering strings with spaces (or just tags separated by space not comma, as users might attempt) in the "tags" field, it has unexpected effects. If there's a nonempty description, the tag string will be appended to it. If there is no description, the tag string is appended to the category name, and if the category isn't filled in either, the tag string will *become* the category.

Well actually here, by observing carefully what happens in the cmdline field when I do this, I could foresee that this wasn't going to work as I expected. Yet I believe it will cause confusion. Syntax checking and not accepting invalid fields might be in order here, too.

ederag:
@mwilck Justified concerns.

This PR is part of a larger plan.
When it is finished, there will be a much better separation of concerns,
as well as more possibilities,
while keeping existing usages safe.

That's why the header said
> To fix #280 and #423 some changes will be required in db.py (working on it).

Actually it's done and just waiting for this PR.

>  being able to fully control hamster with the single command line is absolutely a key feature

If you accept to stick to current limitations, this will still be possible, no change.

Yet parser limitations have nothing to do with gui (https://github.com/projecthamster/hamster/issues/280#issue-125937317).
> Maybe there is a purpose for that in CLI version but in GUI it is confusing.

More importantly from a developer point of view:
The current stage makes any parser change scary (and some changes might not be deliberate).
After a few changes, old facts in the database can be impossible to edit or even see (#423).
Legacy code is really good, but that part should be improved IMO.

The plan is to make database and the separate fields "*verbatim*", without any parsing.
Parsing is confined to command line and gui cmdline.
(this is ready to be pushed)

What would be important to have is an indicator
about whether the fact entered through separate fields is parseable or not.
(probably through `Fact.parse(repr(fact))` == `fact`).

What should be this indicator ?
Please open an issue after this one is merged, to discuss that.

Hiding the calendar is almost finished. Stay tuned.

ederag:
Done, calendar is hidden by default,
but date is shown, including week-day.
This should hopefully prevent most mistakes.
Screenshot_20191029_134925

The `label fill` is on, yet the label itself has to be clicked,
the clickable area does not extend to the full width.
That's why I left the date label also when the calendar is shown.

ederag:
Forgot to push the `label fill` stuff.
@mwilck 
> GtkExpander widgets can be set to occupy the whole width of the parent widget

Did nothing here. Is it my system (gtk-3.22) ?

That can be fixed later. Let's merge this one tomorrow.

GeraldJansen:
Hurray!

ederag:
Merged, thanks for the feedback !

Reminders:
- [ ] Database access should be verbatim to fully exploit the separate fields: https://github.com/projecthamster/hamster/pull/461#issuecomment-547386843.
- [ ] A warning sign will be needed to indicate verbatim entries not compatible
with the cmdline limitations.
- [ ] Empty categories do not work yet (but they never did even in v1).
- [ ] We'll see [later](https://github.com/projecthamster/hamster/pull/461#issuecomment-546880690) if more completions can be added to the cmdline.

mwilck:
@ederag, many thanks for your late improvements.

> Database access should be verbatim to fully exploit the separate fields

I'm not happy about this part of the update, [rationale](https://github.com/projecthamster/hamster/pull/461#issuecomment-547326960). I believe that adding the ability to enter content to the data base that can't be parsed was a move in the wrong direction. The intention to improve the parser (#465) is good, but maybe we should first have discussed what the parser could possibly do. The current rules (no quoting, and spaces allowed in tokens) make it essentially impossible to write a parser that could be fully compatible to separate fields input.

Anyway, it's how it is now. Let's hope that users don't shoot themselves in their feet with the separate entry fields. I for one am not going to use them anyway.

ederag:
@mwilck Changing the parser is no small feat,
and do not fix several bug that will need separate fields anyway.
Your concern has already been taken seriously, and is tracked in #467.

GeraldJansen:
Separate input fields seems to be important to some users and will ease the transition from v1, so fine. I may  even use them occasionally. 

What I really dislike is the whole idea of verbatim input fields allowing input that cannot be edited in the cmdline (or input from the CLI), and then having to visually signal non-standard inputs. That seems pretty messy to me. Releasing this and PR 466 will allow input like this:
```
./hamster-cli list
Start | End | Duration | Activity                               | Category
-----------------------------------------------------------------------------
10:51 |     | 0h 46min | x@y  bis #123, some text #tag@Unsorted | zzz     
-----------------------------------------------------------------------------
```
Once these changes are released into the wild, it will be too late to start discussing future parsing improvements.

I suggest a more cautious approach for the upcoming release. In the activity entry field, only allow '#' as the very first character (as per the original request), and disallow ',' and '@'. This would require a minimal change to the cmdline parser to allow the initial '#'.

mwilck:
@GeraldJansen, full ACK.

CendioOssman:
A bit late to the party, but for future reference. :)

> Question is, how often do you enter a brand new activity, with respect to reuse an existing activity ?
> (what ratio ?)

The ratio would be close to infinite in favour of existing categories. In fact, for me it would be a feature if new categories could **not** be entered that way and have to be explicitly set up.

mwilck:
> The ratio would be close to infinite in favour of existing categories. In fact, for me it would be a feature if new categories could **not** be entered that way and have to be explicitly set up.

Good point. This holds for most long-time users I guess. But "explicitly set up" should not mean having to open the "Tracking Settings/Categories and Tags" dialog. That would be way too explicit for my taste.

ederag:
@CendioOssman 
> The ratio would be close to infinite in favour of existing categories

Thanks for the confirmation.

@GeraldJansen, @mwilck Granted, #465 has not been demontrated yet, 
so taking into account your concerns, 
facts are not allowed to be saved from the gui if they can not be parsed back (PR #476).

Issue #465, opened 2019-10-25
cmdline entry format improvements


ederag:
There have been requests for a more permissive parser (#280, #423)
and different suggestions such as
```activity name[@category[, description] [#tag1] ... [#tagn]]```
in https://github.com/projecthamster/hamster/issues/438#issuecomment-529860808, or
```name[@Category] [#tag1] [#tag2] ... [description]```
in https://github.com/projecthamster/hamster/issues/438#issuecomment-546314277.

My current opinion is that there is no way to have a smart parser
using heuristics suitable for almost any input.
This is even more difficult as this parser should keep 
the currently satisfied users experience intact.

Both the above suggestions do not keep compatibility.
The first one because it would not be possible to enter `activity, description` without category.
The second one because description would now be at the end.
Another issue with the second one is that it would prevent entering a description starting with `#`.

So the plan could be to make all separators configurable, such as `,,` before description.
Regular expressions could be allowed.
The defaults would not change, so most people would not even notice.

mwilck:
>   The second one because description would now be at the end.

Not so bad because the current format has never been thoroughly described :-) Also, the description parsing after comma doesn't seem to work at all in hamster 2.2.2, so I don't see a regression there.

We could also define the format like this:

    name[@Category] [description and tags]

where "description and tags" would be just free-form input from which hamster would pick all #words prefixed with #hash as #tags. That would include tags at the beginning, at the end, or actually anywhere after the first word.

Thus:

    8:00 - 9:30 #boss@meeting In this meeting I asked for higher #salary and more #days_off

would define an activity "#boss" in category "meeting" with description "*In this meeting I asked for higher salary and more days_off*" and tags "salary" and "days_off".

This should be parseable just fine.

In general, I'd recommend against making this configurable. This is GNOME - avoid possibly confusing configuration options.

GeraldJansen:
A nitpick, but how about `#boss, I'm asking for higher #salary`?
It's simpler if there is a rule that an @category is mandatory if you want to add tags and/or description, hence:
`activity name[@category[, description] [#tag1 #tag2 ...]]`

I suggest choosing something that is similar to past practice (but not necessarily 100% compatible), convenient for entry and simple to define and parse. Define the behaviour in a wiki page as the standard, then stick to it. It will be impossible to please everyone.

I am also not in favor of the further complexity of configurable separators.

ederag:
> I am also not in favor of the further complexity of configurable separators.

[...]

PR #466, opened 2019-10-30
Verbatim database read/write

ederag:
With this PR database read/write is done verbatim, without any parsing involved.
This makes a clear separation of concerns.
It partially solves #280 and #423,
although that requires renouncing to the quickness of the cmdline input, until #465 is solved.

This change should not affect the existing dbus interface (old functions have been kept).
The new `AddFactVerbatim` is strongly encouraged if the caller has separate fields.
The old `AddFact` will remain so that passing a single string will remain possible,
to be interpreted by `Fact.parse` on the hamster side.

Unsorted (no category) facts are not supported yet (forthcoming PR almost ready).

### DIFF
--- a/src/hamster-service
+++ b/src/hamster-service
@@ -168,6 +168,14 @@ class Storage(db.Storage, dbus.service.Object):
         if end_time:
             end_time = dt.datetime.utcfromtimestamp(end_time)
         return self.update_fact(fact_id, fact, start_time, end_time, temporary) or 0
+    
+    
+    @dbus.service.method("org.gnome.Hamster",
+                         in_signature='i{}'.format(fact_signature),
+                         out_signature='i')
+    def UpdateFactVerbatim(self, fact_id, dbus_fact):
+        fact = from_dbus_fact(dbus_fact)
+        return self.update_fact(fact_id, fact) or 0
 
 
     @dbus.service.method("org.gnome.Hamster", in_signature='i')
--- a/src/hamster-service
+++ b/src/hamster-service
@@ -18,7 +18,7 @@ from hamster.lib.dbus import (
     dbus,
     fact_signature,
     from_dbus_fact,
-    to_dbus_fact_verbatim
+    to_dbus_fact
 )
 
 logger = default_logger(__file__)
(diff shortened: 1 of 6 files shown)


ederag:
@GeraldJansen Could it be that `hamster-service` from a previous version is installed ?
In this case, launching `hamster-cli` queries dbus through `client.py`,
that automatically relaunches the installed `hamster-service`, 
which lacks the new `AddFactVerbatim` method.

Could you try to launch the new hamster-service first ?
```bash
pkill -ef hamster.*service
./hamster-service
./hamster-cli track a long, hard furrow
```
If a service is already found running, it is used.

ederag:
By the way, using your example gives an "a" in activity, the rest is lost.
But that was already the case previously (checked with 5f782f6fecb0c46db2332ce6c16bf287a34379dd).
`Fact.parse` will be reworked to make it usable for command line parsing, after v3.0.

GeraldJansen:
> Could it be that hamster-service from a previous version is installed ?

Right, of course. Sorry for the noise.

OT.  Now getting this error message every minute on update:
`hamster.lib.desktop - ERROR - Error while refreshing: 'Fact' object is not subscriptable`

On a plus note, I don't find any noticeable slowdown in auto-completion from the cmdline. Its still very snappy for me.

ederag:
Thanks, this one slipped under the radar.
1aefbba2d5ef9fcc5e6ba1d8cf96069893a4aaea is a blind guess. Is it silent now ?
Need to find who is calling `desktop.py`...

GeraldJansen:
> 1aefbba is a blind guess. Is it silent now ?
Yes. Good guess.

GeraldJansen:
A small bug (edge case?): if I update an activity that has a category, clear the category and save, the saved fact still has the old category, not Unsorted.

ederag:
Indeed, that's because after clear the fact has no category.
> Unsorted (no category) facts are not supported yet (forthcoming PR almost ready).

That needs a 3 lines only fix, ready, but unfortunately it is slightly breaking the dbus interface:
`GetFact` and `GetFacts` would return an empty string instead of "Unsorted".
You may try this [fix-unsorted](https://github.com/ederag/hamster/tree/fix-unsorted) branch,
but please do not base anything on it, as more rebase are expected.
What do you think ?

GeraldJansen:
No further comment. As an aside, I've always found the @Unsorted presence in the cmdline strange and unsightly. I always assign a real category.

PR #473, opened 2019-11-05
Fix unsorted

ederag:
This is slightly changing the DBus interface,
as `GetFact` now return an empty string instead of `"Unsorted"` for an empty category.
Unsorted never worked in v1 anyway, so it should not matter much.


GeraldJansen:
I bit late, but thanks a lot for this. That "@Unsorted"  bugged me and I'm happy to see it gone.

PR #475, opened 2019-11-16
Accept activities starting with '#'

ederag:
Solves #423.

Note: explicitly forbid comma in tags at parse time
It would be failing anyway during edition of the fact,
because tags entries are based on a comma separated list of tags.


ederag:
To superseded by #482

Issue #123, opened 2013-12-03
Resume last selected activity on unlock or interrupted idle state


audriusrudalevicius:
Add feature to resume last selected activity on unlock or interrupted idle state

PR #476, opened 2019-11-16
Forbid unparseable facts

ederag:
Make the save button active only if the fact entered can be parsed (and thus edited) from the cmdline.
This hopefully answers https://github.com/projecthamster/hamster/pull/461#issuecomment-549302989 favorably, while waiting for #465.

The first two commit are from PR #475, only a544d5e is added by this PR.

### DIFF
--- a/src/hamster/lib/__init__.py
+++ b/src/hamster/lib/__init__.py
@@ -337,15 +337,21 @@ def parse_fact(text, phase=None, res=None, date=None):
             if not m:
                 break
             tag = m.group(1)
-            tags.append(tag)
             # strip the matched string (including #)
+            backup_text = remaining_text
             remaining_text = remaining_text[:m.start()]
+            # empty remaining text means that activity is starting with a '#'
+            if remaining_text:
+                tags.append(tag)
+            else:
+                remaining_text = backup_text
+                break
         # put tags back in input order
         res["tags"] = list(reversed(tags))
         return parse_fact(remaining_text, "activity", res, date)
 
     if "activity" in phases:
-        activity = re.split("[@|#|,]", text, 1)[0]
+        activity = re.split("[@|,]", text, 1)[0]
         if looks_like_time(activity):
             # want meaningful activities
             return res
--- a/src/hamster/lib/__init__.py
+++ b/src/hamster/lib/__init__.py
@@ -23,7 +23,7 @@ TIME_FMT = "%H:%M"
 tag_re = re.compile(r"""
     [\s,]*     # any spaces or commas (or nothing)
     \#          # hash character
-    ([^#\s]+)  # the tag (anything but # or spaces)
+    ([^#\s,]+)  # the tag (anything but #, spaces or comma)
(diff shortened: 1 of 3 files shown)


GeraldJansen:
Thanks! Yes this solves [#461 (comment)](https://github.com/projecthamster/hamster/pull/461#issuecomment-549302989) reasonably well, as well as the two issues closed by the previous two commits. 

I noticed that the parser changes allow "#" in any position in the activity name, in addition to the first, which is fine with me. Playing with this a bit, I noticed some quirky behaviour. For example, typing ` #123 and #456@bugs, description` slowly in the cmdline, ` 456@bugs` is considered a tag up until the ', d' is entered and then flips to be part of the description and category. I guess this is a corner case that can be addressed later under PR #475.

(ps. very glad you're back!)

ederag:
@GeraldJansen 
> 56@bugs is considered a tag up until the ', d' is entered and then flips to be part of the description and category

Thanks for the feedback, glad you're still there !
That kind of interpretation change seems inevitable since
`activity #456`  was interpreted as an activity with `456` tag,
and we want to be fully backward compatible.

Just noticed: adding a comma in the middle of a tag in the gui cmdline wreaks havoc. :disappointed:

src/hamster/lib/__init__.py
```diff
@@ -23,7 +23,7 @@
 tag_re = re.compile(r"""
     [\s,]*     # any spaces or commas (or nothing)
     \#          # hash character
-    ([^#\s]+)  # the tag (anything but # or spaces)
+    ([^#\s,]+)  # the tag (anything but #, spaces or comma)
```
GeraldJansen: This change would fix the strange behaviour with `#456@cat`:
```
    ([^#@\s,]+)  # the tag (anything but #, @ederag , spaces or comma)
```

GeraldJansen: `([^#@\s,]+)  # the tag (anything but #, @ , spaces or comma)`

ederag: Wouldn't that forbid '@' in tags and be a regression ?
IMO the behavior is certainly puzzling at first sight, but clearly understandable
if the rules can be kept simple.

GeraldJansen: Ah, so we also need to allow @ in the middle of tags. I never would have imagined that. Okay.

GeraldJansen:
Looking at the Input help page (eg. [here](https://geraldjansen.github.io/hamster-doc/input.html)), it seems that "#tag with spaces" is explicitly mentioned. This no longer works with this PR (or maybe earlier). For cmdline input `abc@cat, desc #tag with spaces` the description becomes "desc #tag with spaces" and there are no tags. A regression?

ederag:
Perfect, that used to work in v2.0 (c3e5fb761c88fdecfd1566cac8b6836228a27cce)
it would have been a pity not to allow spaces in tags as well :smile:

ederag:
superseded by #481

PR #481, opened 2019-11-19
Forbid unparseable facts - take2

ederag:
Make the save button active only if the fact entered can be parsed (and thus edited) from the cmdline.
This hopefully answers https://github.com/projecthamster/hamster/pull/461#issuecomment-549302989 favorably, while waiting for #465.

[This is the last commit of #476, which will be closed]


GeraldJansen:
LGTM

PR #483, opened 2019-11-20
Automatic determination of version string

GeraldJansen:
Use `git describe --tags --always --dirty=+` to automatically determine a Hamster version string. The version string is set by the waf wscript for installed versions or at runtime if running uninstalled. For example,
```
[src]$ ./hamster-cli version
v2.2.2-316-g24c88ae (uninstalled)
```
where the current revision is 24c88ae and it is 316 commits after tag v2.2.2. 
(Also works for `hamster about`.)


ederag:
Same here, this looks good (hence the hooray tag). Please reopen ?
Sorry, I have just been dragged into too many potentially breaking changes 
to sort them out as early as possible. [some are not pushed yet]
The idea was that this PR can be reviewed with a free mind, 
and merged safely even close to the release.
i apologize for not leaving a comment here previously.

GeraldJansen:
Same as for #484.The commit 24c88ae is minimal so you can easily recover the changes if and when you wish.

ederag:
OK, I'd be glad to see you around from time to time,
and I wish you the best for your new priorities !

ederag:
Github is not smart enough. 
This PR still looks closed, even if all its commits have been incorporated into the #528 followup.

Issue #238, opened 2015-05-19
Unable to remove tags


Forage:
Sometimes you get unwanted tags because of a typo, especially because tags don't get auto-completed in the GNOME extension (bug #114 and #12), or simply because you change your mind afterwards. The result is tags you can't remove any more, despite an option to remove tags from the complete field. Removing tags from this field in the preferences still presents all of them again when you return to the list.

Steps to reproduce:
1. Add activity through the extension "reading e-mail@work, #pita"
2. Add activity through the extension "poo@work, #relieve"
3. Open the preferences through "Tracker settings" option in the extension and go to the "Categories and Tags" tab
4. Remove "pita" from the auto-complete list
5. Close the preferences and repeat step 3.

Result:
1. The tag "pita" is still present in the auto-complete list

This happens for tags that are being used for existing activities as well as non-used tags because you changed/fixed them by altering an existing activity.

OS: Ubuntu 14.04
Hamster: 3.0really1.03.3+git2370-0dylanmccall1~ubuntu14.04.1 from https://launchpad.net/~dylanmccall/+archive/ubuntu/hamster-time-tracker-git-stable

Let me know if you need any more information.

Johann-Tree:
I have the same issue here, very annoying: It created several wrong tags from wrong auto-completion. Until this is fixed I directly edit the database via SQLite Manager (e.g. for Firefox https://addons.mozilla.org/firefox/addon/sqlite-manager ). This works ok.

orschiro:
For Ubuntu Linux users I recommend http://sqlitebrowser.org/.

ederag:
In current master (d86e07fc6550f53713917d6308584400c8bdbbc3) the tags can be removed, provided they do not appear in an existing fact.
So totally wrong tags can be removed (just fix the wrong fact first).

But still it should be possible to remove a tag, whether it is present in a fact or not.

ederag:
Should be fully fixed by #487 (thanks @GeraldJansen !).

PR #487, opened 2019-11-29
fix bug preventing removal of autocompletion tags

GeraldJansen:
This fixes a tiny bug in concerning tag autocomplete flags (1 vs. 'true') in the sqlite database, fixed here by checking for not 'false'. See #238.

### DIFF
--- a/src/hamster/storage/db.py
+++ b/src/hamster/storage/db.py
@@ -206,7 +206,7 @@ class Storage(storage.Storage):
         gone = self.fetchall(query, tags)
 
         to_delete = [str(tag["id"]) for tag in gone if tag["occurences"] == 0]
-        to_uncomplete = [str(tag["id"]) for tag in gone if tag["occurences"] > 0 and tag["autocomplete"] == "true"]
+        to_uncomplete = [str(tag["id"]) for tag in gone if tag["occurences"] > 0 and tag["autocomplete"] != "false"]
 
         if to_delete:
             self.execute("delete from tags where id in (%s)" % ", ".join(to_delete))
--- a/src/hamster/storage/db.py
+++ b/src/hamster/storage/db.py
@@ -169,7 +169,7 @@ class Storage(storage.Storage):
         changes = False
 
         # check if any of tags needs resurrection
-        set_complete = [str(tag["id"]) for tag in db_tags if tag["autocomplete"] == "false"]
+        set_complete = [str(tag["id"]) for tag in db_tags if tag["autocomplete"] in (0, "false")]
         if set_complete:
             changes = True
             self.execute("update tags set autocomplete='true' where id in (%s)" % ", ".join(set_complete))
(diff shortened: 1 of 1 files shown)


ederag:
Thanks for providing a fix.
Checking for non-equality seems less safe than for equality.
(equality matches one case, while non-equality matches any other type instances)

Checking for `"true" or 1` would not help much,
as this feels like something weird is going on.
What triggered the change from true/false to 1 in the first place ?

GeraldJansen:
The code is inconsistent. New tags created from adding facts get 1. Recovered tags that were 'false' get set to 'true' (in preferences iirc). At any rate, in existing databases, both 1 and 'true' may be in the DB, whereas on the negative side there is only 'false'. I found this check for autocomplete != 'false' at another place in the code.

ederag:
> The code is inconsistent. New tags created from adding facts get 1.

This does seem to be explicitly set. Is that a default value ? Yes:

Using `sqlitebrowser ~/.local/share/hamster-applet/hamster.db`, in `database structure`,
for tags:
```sql
CREATE TABLE tags (id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, 
                   name TEXT NOT NULL, autocomplete BOOL DEFAULT true)
```
so `autocomplete` is expected to be boolean, not text.
Note: `BOOL` is not defined in sqlite, [only `BOOLEAN` is](https://www.sqlite.org/datatype3.html#affname).
Besides, python `sqlite3` has [no adapter for booleans](https://docs.python.org/3.7/library/sqlite3.html?highlight=sqlite3#introduction).

A `CHECK` would have prevented this, see
https://stackoverflow.com/a/22186315/3565696
http://www.sqlite.org/lang_createtable.html#ckconst
No; that would break the database compatibility.

> I found this check for autocomplete != 'false' at another place in the code.

Indeed, found one in `__get_tags` but that is read only, not as risky as writing.

Would replacing all the settings to `"true"`/`"false" to be `1`/`0` right now,
break previous versions ?
(using the same database with previous versions is part of the compatibility constraint) 
Not for this line that was [not changed in ten 10 years](https://github.com/projecthamster/hamster/blame/46954af249f393f6baebe0ff3934bb1331d2ae8e/src/hamster/storage/db.py#L159)
https://github.com/projecthamster/hamster/blob/46954af249f393f6baebe0ff3934bb1331d2ae8e/src/hamster/storage/db.py#L159
But this one would be affected:
https://github.com/projecthamster/hamster/blob/46954af249f393f6baebe0ff3934bb1331d2ae8e/src/hamster/storage/db.py#L172

So what about replacing the conditionals 
`== "true"` or `!= "false"` 
to 
`in (1, "true")`
and `== "false"` to `in (0, "false")`
both in the sql queries and in the python conditionals,
to get something safe and backward compatible,
and drop the "true"/"false" later ?

GeraldJansen:
Too many questions. The simple bug fix is good enough for me. Please do as you see fit with this PR.

ederag:
Sorry, understood :smile: 
and you are right that your fix could be used.
The last proposition just feels even safer.
The rest were mainly notes taken while investigating.
Only the last question really matters.

GeraldJansen:
If you prefer to put `in (1, 'true')` and `in (0, 'false')` everywhere, please do so. No offence taken. I prefer to not discuss this further.

PR #490, opened 2019-12-10
restore AddFact compatibility

ederag:
DBus compatibility was lost, as reported by @jmberg in the shell-extension:
https://github.com/projecthamster/hamster-shell-extension/pull/316#issuecomment-564145295

This is a blind tentative to fix it.
(shell extension does not work here, and an earlier attempt to install it in an ubuntu virtualbox failed)


jmberg:
Looks right, I'll test it in a minute.

jmberg:
Yes, works for me with the gnome-shell-extension (that includes that other pull request)

jmberg:
thanks :-)

ederag:
Thanks a lot for the feedback !
This was pulled in #482 as well, if you want to try it
(still need polishing, but hopefully better than master)

Issue #418, opened 2019-06-17
Snap package for hamster


extraymond:
Hi! Guys, I'm trying to build [hamster in snap](https://forum.snapcraft.io/t/need-help-dbus-activation-for-project-hamster-snap/11885) right now, I believe I'm half way there.

## Motivation

My motivation for doing so is as follows:
1. Snap is a great way to maintain stable run-time environment so app will continue to work regardless of distro versions. It support various build systems too, waf is automatically handled.
2. Snaps have some nice update mechanisms, we can easily switch between channels as desired. It have some safety mechanisms that perform auto rollback when newer version crashed.
3. Combining the two, I think it might give hamster a greater chance to reach it's audience.

So here I am trying to reach out and see if anything clicks.

---

## Problems

However there are some problems regarding invoking services with dbus.

For the manual build version, the hamster launch file with automatically invoke the hamster-service and hamster-windows-service if it's not started, but on the snap service I can't find a way to do so. So currently I have to manually start the two service before launching the GUI. Which works!

I would really appreciate any help regarding to setup dbus service so that it can be invoked by dbus name.

Here's the snapcraft.yaml if anyone is interested:

```yaml

name: hamster-snap # you probably want to 'snapcraft register <name>'
base: core18 # the base snap is the execution environment for this snap
version: "0.1" # just for humans, typically '1.2+git' or '1.3.2'
summary: snap for project hamster # 79 char long summary
description: |
  First attempt for creating a snap for project hamster.

grade: devel # must be 'stable' to release into candidate/stable channels
confinement: devmode # use 'strict' once you have the right plugs and slots

slots:
  dbus-hamster:
    interface: dbus
    bus: session
    name: org.gnome.Hamster

plugs:
  gnome-3-28-1804:
    interface: content
    target: $SNAP/gnome-platform
    default-provider: gnome-3-28-1804
  gtk-3-themes:
    interface: content
    target: $SNAP/data-dir/themes
    default-provider: gtk-common-themes
  icon-themes:
    interface: content
    target: $SNAP/data-dir/icons
    default-provider: gtk-common-themes
  sound-themes:
    interface: content
    target: $SNAP/data-dir/sounds
    default-provider: gtk-common-themes

apps:
  hamster-service:
    command: desktop-launch ${SNAP}/usr/lib/hamster-time-tracker/hamster-service
    plugs:
      - desktop
      - desktop-legacy
      - gsettings
      - home
    environment:
[...]

PR #495, opened 2019-12-20
Streamline installation instructions

ederag:
As adequately noted in https://github.com/projecthamster/hamster/issues/418#issuecomment-567891068, the installation instructions are not clear enough.

@tchama Is [this proposition](https://github.com/projecthamster/hamster/blob/98b8de3a92ad72b1d5f4aab395bfc5db77b9e405/README.md) better ? Thanks for helping ironing out the installations instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,17 +33,18 @@ months (i.e. late 2019).
 
 ## Installation
 
-You can use the usually stable `master` or [download stable releases](https://github.com/projecthamster/hamster/releases).
+#### Backup database
 
-If you upgraded from an existing installation make sure to kill the running
-daemons:
+This legacy hamster should be stable, and keep database compatibility with previous versions.  
+It should be possible to try a new version and smoothly roll back to the previous version if preferred.  
+Nevertheless, things can always go wrong. It is strongly advised to backup the database before any version change !
+
+##### Locate the latest db
 
 ```bash
-pkill -f hamster-service
-pkill -f hamster-windows-service
-# check (should be empty)
-pgrep -af hamster
+ls --reverse -clt ~/.local/share/hamster*/*.db
 ```
+Backup the last file in the list. 
 
 #### Dependencies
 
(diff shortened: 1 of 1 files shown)


tchama:
I was able to follow the instructions of the main module, but then when I launch hamster, I get an error that I understand, such as the fact that one or more components are missing in my installation.

On the page https://github.com/projecthamster there are 6 pinned repositories. The main obstacle is to know which module I need, and what they are used for. 
I have the impression that there is a lack of an overview of these modules, to know what they are for, which ones are needed, in which order they should be compiled, etc.

ederag:
This repository is monolithic (single module),
no need for the `hamster-*` modules that are part of the separate "rewrite" project (currently stalled).

Could you please paste the error message ?

tchama:
ok. Here is my [hamster_install_log.txt](https://github.com/projecthamster/hamster/files/3991351/hamster_install_log.txt).

After that I just tried to launch hamster
```
mat @ swan  ~
└─ $ ▶ hamster
  File "/usr/bin/hamster", line 30, in <module>
    from hamster import client, reports
ImportError: cannot import name 'client' from 'hamster' (/home/mat/.local/lib/python3.7/site-packages/hamster/__init__.py)
```
Thanks for trying to help me install it.

ederag:
Thanks for the full log. It looks fine, installation is in the standard `/usr/lib/python3/dist-packages/`.
But at launch time, the import is looking into
`/home/mat/.local/lib/python3.7/site-packages/hamster`.
That is probably a remnant of a former installation attempt (maybe `pip install hamster-sqlite` ?).
You might try to move `/home/mat/.local/lib/python3.7/site-packages/hamster` out of the way.

tchama:
> That is probably a remnant of a former installation attempt (maybe `pip install hamster-sqlite` ?)

That was it. I removed `hamster-sqlite` from the python packets. Hamster's launching now. I still have some hamster packets left in python. Should I take those out, too?
```
└─ $ ▶ pip3 list | grep hamster
hamster-cli            0.12.0       
hamster-gtk            0.11.0       
hamster-lib            0.13.2       
hamsterlib             0.1.0    
```
Thanks for the push.

ederag:
Glad it works !
All 4 packages are from the hamster rewrite, and could be removed, as far as [hamster](https://github.com/projecthamster/hamster) is concerned.
But there should not be any conflict either, as they use different package names and DBus interface.

@elbenfreund [hamsterlib](https://pypi.org/project/hamsterlib/) looks like a failed attempt, superseeded by [hamster-lib](https://pypi.org/project/hamster-lib/) ?

PR #496, opened 2019-12-23
Initial snapcraft attempt.

extraymond:
Hi! After some trial and error, this is my attempt to create a hamster snap with the least possible changes regards the code base.

A introduction to this snap:
1. Reuse existing .desktop file.
2. Change /usr/share path to $SNAP/usr/share to satisfy relative path under snap.
3. Use gnome-3-28 extension to provide a stable gnome environment and reduce snap size.

Current limitation and noticeable changes:
1. Snap confinement set to **devmode** instead of **stict**. To my understanding. This is due to gconf related issues. Since changing gconf to gsettings is underwent. I will revisit this part after so I won't break too much stuff.
2. Manual launch script to start the two services before the GUI, so it won' crash. This is due to dbus-activation not currently possible for snap. There's upstream discussion between snap and dbus. Will update if possible.
3. Change defs.py to defs_comp.py, and create a defs.py to proxy the defs_comp.py. Alter the path with os.environ['SNAP'] if detected to be launched by snap.

A included snap to play with. You can try with snap install hamster-snap*.snap --devmode


extraymond:
Anyone interested in testing. Here's a prebuild snap.
[hamster snap](https://www.dropbox.com/s/6lvn1mhmf66x7s3/hamster-snap_0.1_amd64.snap?dl=0)

ederag:
Thanks for your proposition !
Having a snap package would be good, indeed.

This could belong to a separate repo (maybe under you github account), say `hamster-snap` ?
That would greatly split the maintenance responsibility.
Moreover, you would be able to develop and release at will.

Then we'd add an advertisement section to the README.md.

I'm leaving this PR open until then.

*Side question:*
why rename `defs.py` to `defs_comp.py` ?

extraymond:
> Thanks for your proposition !
> Having a snap package would be good, indeed.
> 
> This could belong to a separate repo (maybe under you github account), say `hamster-snap` ?
> That would split the maintenance responsibility.
> Moreover, you would be able to develop and release at will.
> 
> Then we'd add an advertisement section to the README.md.
> 
> I'm leaving this PR open until then.
> 
> _Side question:_
> why rename `defs.py` to `defs_comp.py` ?

About defs.py:

Because typically, when /usr/share points to the host /usr/share, so *.ui files and files needed can be imported just by relative path. But in a snap, the /usr points to the host outside of the snap environment. Where the files are installed in the /SNAP/app/version/usr/share convention, since defs got imported by multiple places. I think it's easier to just create a proxy module and work with it.

While it's possible to maintain on my own repo, I think it works better for discoverability. And snap have tip/donate feature so if you guys are the maintainer it would go straight to you.

To build a snap, the minimal requirement is to have the snapcraft.yaml. And it can be build using the github integration feature from snapcraft.io, so it can have continuous integration on the fly. And users would be able to receive upgrades as fast as possible. It can also set multiple channel so user can choose to use non-stable version in the future, I think that might help testing too.

Once again, I'm ok to maintain it on my own if you think it's inappropriate right now.

ederag:
> Because typically, when /usr/share points to the host /usr/share, so *.ui files and files needed can be imported just by relative path. But in a snap, the /usr points to the host outside of the snap environment. Where the files are installed in the /SNAP/app/version/usr/share convention, since defs got imported by multiple places. I think it's easier to just create a proxy module and work with it.

This is not clear to me. The `--prefix` option might help ?
```bash 
./waf configure build --prefix=/SNAP/app/version/usr
```


> While it's possible to maintain on my own repo, I think it works better for discoverability. 

For discoverability, your package could eventually move to [projecthamster](https://github.com/projecthamster/).
But only @elbenfreund could do that at the moment.
Meanwhile, a good advertisement on the [README](https://github.com/projecthamster/hamster) would be efficient, IMO.

>And snap have tip/donate feature so if you guys are the maintainer it would go straight to you.

Thanks, but that would be inappropriate, as previous developers contributed a lot more than I did.
That kind of donations should be redirected to a FreeSoftware association,
which can be done in a separate repo as well :wink:.

> github integration feature

A `hamster > hamster-snap` snapcraft.io chain is probably feasible with github integration ?

> Once again, I'm ok to maintain it on my own if you think it's inappropriate right now.

Thank you for your comprehension, and the useful addition to the hamster ecosystem !
Please ping here when the repo is set up.

extraymond:
That sounds reasonable. I will this in my own repo.

About the defs.py part.
Snaps are all about confinement, so without our explicit agree, it will not be able to break out the environment if confined.

So although the waf pluging tells it to installed to /usr/, it's actually contained in the snap environment itself, which is /snap/app/version/usr/..., all the assets are mounted in that path, which works like a container. But when executing the snap. It's run by the user from $HOME, so it will try to execute the app from the user's standpoint. In that case, /usr/var/lib is an absolute string which points to /usr/var/lib for the user. But since the assets are mounted in the contained snap, it's required to give the correct path to the method.

Another way to handle this is to use os.path.realpath, but since I would like a minimal changes to the code base as possible, I think proxying the path's are easier.

This is inside the snap.
2019-12-24 02-06-51 的螢幕擷圖
Which is configured exactly what waf told it to.

ederag:
One thought: for your hamster-snap package, please do not fork this repo, but rather
point the `source` keyword to the [hamster](https://github.com/projecthamster/hamster) remote Git repository, 
as explained in https://snapcraft.io/docs/python-apps.

> Another way to handle this is to use os.path.realpath

Where would it be required ?
If this improves the general situation (e.g. make hamster more robust), 
small code changes might be acceptable.

extraymond:
> One thought: for your hamster-snap package, please do not fork this repo, but rather
> point the `source` keyword to the [hamster](https://github.com/projecthamster/hamster) remote Git repository,
> as explained in https://snapcraft.io/docs/python-apps.
> 
> > Another way to handle this is to use os.path.realpath
> 
> Where would it be required ?
> If this improves the general situation (e.g. make hamster more robust),
> small code changes might be acceptable.

Understood! If the defs.py issue can be solved, I would just need to point to master and patch the launcher locally.

After some research I think os.path.realpath doesn't help much unfortunately. If a program is in snap environment, it will show $SNAP in os.environ. I only need the to prefix that to make loadUI and etc work.

extraymond:
Update!!! Turns out we don't need to patch defs.py!!!

We can just reroute the specific path toward the $SNAP one.
[Layout](https://snapcraft.io/docs/snap-layouts)

Issue #166, opened 2014-06-19
Changing time zone while tracking results in incorrect time


rbuch:
If one changes the system time zone while tracking is ongoing, the timing data is incorrect.  The time reported is incorrect by the delta between the two time zones; it appears to be using the raw numerical time instead of taking the time zone into account when calculating elapsed time.

PR #510, opened 2020-01-12
Hamster datetime

ederag:
Introduce `hamster/lib/datetime.py`.
This is the first step towards fixing #166.
For now times are still "naive", 
but this PR should allow us to smoothly shift to "timezone-aware" later on.
We also now have full control on the python/sql conversions
(does not change anything yet, but creates opportunities).

Bonus: code seems a bit clearer with parsers close to their objects.
The ephemeral `parsing.py` will soon be removed.

Some `stuff.py` code was transferred to the new `datetime.py`, but compatibility is retained for a while.
Transfer can be finished later.

### DIFF
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -148,7 +148,7 @@ class HamsterClient(object):
         fact = Fact.parse(" ".join(args), range_pos="tail")
         if fact.start_time is None:
             fact.start_time = dt.datetime.now()
-        self.storage.check_fact(fact, default_day=dt.today())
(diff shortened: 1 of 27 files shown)


ederag:
`Fact.start_time` and `Fact.end_time` are still there, but to be replaced later by 
the new `Fact.range.start` and `Fact.range.end`.
`Fact.serialized_range` moved to `Range.format`.
This a bit cleaner, removes ambiguities such as "is `start_time` a `time` or a `datetime` ?",
and will be a bit clearer for future sparse facts
(facts contained in a range, but with total duration < range.end - range.start).

PR #512, opened 2020-01-14
Update snap instruction in README.ME

extraymond:


ederag:
Thanks, but those instruction would be seen by snapcraft users only after facing the issue.
The README points to the [snapcraft page](https://snapcraft.io/hamster-snap) where you already added those tips. 
This is good enough IMO. Or am I missing a point ?

Perhaps not all snap users will know how to kill processes; you might consider to use a link: 
this sentence
> make sure to kill the two background hamster-*-service before launching again, both services will be python processes.

could become
> make sure to [kill the two background hamster-*-service](https://github.com/projecthamster/hamster/blob/master/README.md#kill-hamster-daemons) before launching again.

This step is common indeed, so the section has just been moved to the "installation" top.

extraymond:
That makes sense! Thx for your advice.

Issue #493, opened 2019-12-15
Move idle, external, and similar features elsewhere


ederag:
"External" was never brought back (#271),
and notifications are not working at all (tested on openSUSE 15.1).
There have been related feature requests (#123) and PRs (#118, #259) pending for a long time.

No wonder:
that kind of features is nice to have, but hard to test,
and increase a lot the maintenance burden and responsibility.

All these features can be implemented separately,
talking to the main hamster through dbus, for instance.

Let's remove them from the main repo,
and if there is an interest, eventually delegate that to a new separate project 
(say `hamster-companion`), preferably with another maintainer.
[*I would gladly help to set it up, of course, so please ask here*]

That would even be reusable with the rewrite if it resumes.

ederag:
Removal part done in #494.
Leaving open as an advertisement for the revival part (need help).

ederag:
For JIRA and Redmine integration, [hamster-bridge](https://github.com/kraiz/hamster-bridge) already exists and looks good !
@kraiz the DBus interface is intended to be stable, but we can discuss improvements.

sh-zam:
As a user of this, I will try to fix it :-)

Starting with: notifications are not working

ederag:
@sh-zam that would be good !
In addition to the removed files listed in the PR #494 head comment,
here are some pointers about notifications:
https://wiki.gnome.org/HowDoI/GNotification
https://github.com/projecthamster/unity-indicator/blob/master/hamster-indicator

sh-zam:
Follow up from: PR #521 

If I understand you correctly, you want the notifications to be managed separately (like in a separate service) and it should talk to hamster-service and client over dbus?

ederag:
Exactly.
These features are hard to test, and will need a faster development/review pace, too much for me.

As stated [above](https://github.com/projecthamster/hamster/issues/493#issuecomment-569237187), this is how [hamster-bridge](https://github.com/kraiz/hamster-bridge) works.
@kraiz, @mkorvas, could notifications (PR #521) be added to [hamster-bridge](https://github.com/kraiz/hamster-bridge) e.g. as a [listener](https://github.com/projecthamster/hamster/issues/493#issuecomment-569237187) ?

sh-zam:
Makes sense.

If the repository is fit for hamster-bridge, I'll add it to that.
If not, I'll create a new one for notifications and few other features, I'd love to have :-)

gsobczyk:
[...]

PR #485, opened 2019-11-23
Rename hamster-time-tracker to hamster everywhere

GeraldJansen:
Reduce confusion about hamster / hamster-time-tracker [/ hamster-applet] by renaming everything from hamster-time-tracker to simply hamster (#472). 

Renamed directories (including ~/.local/share/hamster), schemas, desktop files, icons, etc., etc..

(The only place "hamster-time-tracker" is left is in the comments of the translation *.po files.)

PR #521, opened 2020-01-24
Fix notifications

sh-zam:
This _should_ most make the notifications work like it worked in past. (However, it doesn't use the pynotify).

There is one issue, the path to the icon is hardcoded. I am not sure, if we can dynamically access the installed location of hamster or the "future" installation structure of hamster.

CC: issue#493


ederag:
Thanks, this looks very interesting.
But for reasons explained in #493, this should be external (not in hamster itself):
> All these features can be implemented separately,
talking to the main hamster through dbus, for instance.

Please do ask there for any clarifications.

The next weeks free time will be devoted to finalizing v3.0, 
only then will I be able to help setting up an external service.

Meanwhile, please do keep this PR open.
If you merge master into it regularly, people will be able to use it ([instructions](https://github.com/projecthamster/hamster/wiki/How-to-contribute#one-shot-test)),
until the external companion is up and running.

About the icon location, the base path can be accessed in `defs`:
```python
>>> from hamster import defs
>>> defs.DATA_DIR
'/usr/share'
```
But that's about it.  The new icon installation path is set in 
https://github.com/projecthamster/hamster/blob/a572ddcaab1a094ee3bcfa3f8e14ad66ec691582/data/wscript_build#L7-L14

The directory and names changed in #485, 
so beware that it seems to work on your system because of remnants from a previous installation.

There should be a better way though, for instance in the ui, icons can be found by name:
https://github.com/projecthamster/hamster/blob/a572ddcaab1a094ee3bcfa3f8e14ad66ec691582/data/edit_activity.ui#L18

ederag:

ederag:

sh-zam:
I have updated this to the current master and I have updated the way to get icon path, however I am not sure if this is how you're supposed to do it.

src/hamster-cli.py
```diff
@@ -46,7 +46,7 @@
 from hamster.lib import default_logger, stuff
 from hamster.lib import datetime as dt
 from hamster.lib.fact import Fact
-
+from hamster.lib.notifsmanager import notifs_mgr
```
mwilck, line 49: `notifs_mgr` seems to be referenced nowhere in this file. What am I missing?

sh-zam:
Thank you for reviewing!

> I'd like to understand how it relates to the previous code removed in [08eaf77](https://github.com/projecthamster/hamster/commit/08eaf77eb925a1bd0c23bad0bcdb4db7310bfa5c) (`desktop.py`). Have you used that code as a template / reference somehow? 

No, I did not I'm afraid. I wrote it from scratch by looking at the docs :(

> You seem to have implemented some things differently (looking at the actual `Notify()` code for example - missing `transient` and `urgency` properties), could you elaborate a bit on that, either in the commit message or in comments in the code?

I didn't know that such a thing (`transient`) existed. I will update my patch to include it and add a few comments along the way.

PS: Just yesterday, I added code to close notifications automatically. It was an ugly hack. Thanks for mentioning `transient` :-)

src/hamster-cli.py
sh-zam: Must be remnants of my debugging session. I'll remove it, thanks!

PR #531, opened 2020-01-31
Install the service executables in libexecdir

bochecha:
This is the proper directory according to the GNU guidelines:

> The directory for installing executable programs to be run by other
> programs rather than by users.

https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#index-libexecdir

Libexecdir used to be absent from the FHS but it was added in version
3.0 of the standard, which was released in 2015:

> /usr/libexec includes internal binaries that are not intended to be
> executed directly by users or shell scripts. Applications may use a
> single subdirectory under /usr/libexec.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrlibexec

Those executables are the launchers for the background services, and are
not intended to be run by users (e.g the Linux one is run by DBus) which
makes libexecdir the better place to install them.

---

The default as recommanded by GNU is `$(prefix)/libexec`. This is what Linux distributions like Fedora use.

However, other distributions (most notably Debian) mandate in their policy that `libexecdir` must be `libdir`. The second commit in this branch makes it possible for those distributions to override the default.

### DIFF
--- a/org.gnome.hamster.Windows.service.in
+++ b/org.gnome.hamster.Windows.service.in
@@ -1,3 +1,3 @@
 [D-BUS Service]
 Name=org.gnome.Hamster.WindowServer
-Exec=@LIBDIR@/hamster/hamster-windows-service
+Exec=@LIBEXECDIR@/hamster/hamster-windows-service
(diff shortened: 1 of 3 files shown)


ederag:
Thanks for the detailed PR, very well explained, much appreciated !

PR #532, opened 2020-01-31
build: Install the desktop file

bochecha:
Not having this makes it a bit harder to launch the app. :stuck_out_tongue_winking_eye:

### DIFF
--- a/data/wscript_build
+++ b/data/wscript_build
@@ -19,5 +19,6 @@ for filename in ["hamster.desktop"]:
     bld(features = "subst",
         source= "%s.in" % filename,
         target= "%s" % filename,
-        dict = bld.env
+        dict = bld.env,
+        install_path = "${DATADIR}/applications"
        )


ederag:
Indeed, thanks !

Side note: `hamster.desktop` (and hamster-windows-service) should be improved to comply with
https://wiki.gnome.org/HowDoI/DBusApplicationLaunching
Would you by chance also know how to do all this cleanly ?

bochecha:
>  Would you by chance also know how to do all this cleanly ?

I'm not sure. The usual approach is to use GApplication (through GtkApplication) and its `--gapplication-service` command-line option so that the app is started as a background service. Then there are a few trivial things to do with the desktop file, and you're done.

However in the case of Hamster, the DBus service seems to be completely separated from the app if I understood correctly? (they are different processes, different executables, …)

I don't know how to handle such a case. :confused:

ederag:
> The usual approach is to use GApplication (through GtkApplication) and its --gapplication-service command-line option so that the app is started as a background service. Then there are a few trivial things to do with the desktop file, and you're done.

Thanks for the confirmation, that's exactly the goal.

> The DBus service seems to be completely separated from the app if I understood correctly?

Opened #533 to discuss this further.

Reading `hamster-cli.py` again, it is work in progress,
the structure is unclear for a new reader right now.
I'll try this week-end and call for help or review.

ederag:
@bochecha #536 made progress, but now it's stuck. 
Help would be much appreciated, if you have time this week.
Please do tell on #533 if you start working on it, to avoid duplicating efforts.

PR #534, opened 2020-02-01
Fix crash when editing or deleting an activity from overview

sh-zam:
```
dbus.exceptions.DBusException: org.freedesktop.DBus.Python.sqlite3.OperationalError:
  File "/home/sh_zam/workspace/hamster/src/hamster/storage/db.py", line 844, in __remove_index
    self.execute("DELETE FROM fact_index where id in (%s)" % ids)
  File "/home/sh_zam/workspace/hamster/src/hamster/storage/db.py", line 939, in execute
    cur.execute(state, param)
sqlite3.OperationalError: near "(": syntax error``

### DIFF
--- a/src/hamster/client.py
+++ b/src/hamster/client.py
@@ -22,8 +22,10 @@
 import dbus
 import logging
 logger = logging.getLogger(__name__)   # noqa: E402
+import sys
 
 from calendar import timegm
+from distutils.version import LooseVersion
 from gi.repository import GObject as gobject
 from textwrap import dedent
 
@@ -38,6 +40,16 @@ from hamster.lib.dbus import (
     )
 from hamster.lib.fact import Fact, FactError
 from hamster.lib import datetime as dt
+
+
+# bug fixed in dbus-python 1.2.14 (released on 2019-11-25)
+assert not (
+    sys.version_info >= (3, 8)
+    and LooseVersion(dbus.__version__) < LooseVersion("1.2.14")
+    ), """python3.8 changed str(<dbus integers>).
+       That broke hamster (https://github.com/projecthamster/hamster/issues/477).
+       Please upgrade to dbus-python >= 1.2.14.
+    """
 
 
 class Storage(gobject.GObject):
--- a/src/hamster/storage/db.py
+++ b/src/hamster/storage/db.py
@@ -841,7 +841,7 @@ class Storage(storage.Storage):
             return
         ids = ",".join((str(id) for id in ids))
         logger.info("removing fact #{} from index".format(ids))
-        self.execute(["DELETE FROM fact_index where id = ?"], ids)
+        self.execute("DELETE FROM fact_index where id in (%s)" % ids)
 
 
     def __check_index(self, start_date, end_date):
--- a/src/hamster/storage/db.py
+++ b/src/hamster/storage/db.py
@@ -841,7 +841,7 @@ class Storage(storage.Storage):
             return
         ids = ",".join((str(id) for id in ids))
         logger.info("removing fact #{} from index".format(ids))
-        self.execute("DELETE FROM fact_index where id in (%s)" % ids)
+        self.execute(["DELETE FROM fact_index where id = ?"], ids)
 
 
     def __check_index(self, start_date, end_date):

### CODE REVIEW

**Comment by ederag** [comment]
Thanks for the fix, but how can the issue be reproduced ?
Add or remove an activity either from the overview or from the preferences works just fine here.
The database is a very sensitive part, I'd like to investigate further.

**Comment by sh-zam** [comment]
1) Double click the activity in overview page.
2) In Update activity (edit_activity) pop up, change any one thing and click save

Full crash log:
```
$ python src/hamster-cli.py    
  File "/home/sh_zam/workspace/hamster/src/hamster/edit_activity.py", line 371, in on_save_button_clicked
    runtime.storage.update_fact(self.fact_id, self.fact)
  File "/home/sh_zam/workspace/hamster/src/hamster/client.py", line 229, in update_fact
    new_id = self.conn.UpdateFactJSON(fact_id, dbus_fact)
  File "/usr/lib/python3.8/site-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib/python3.8/site-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Python.sqlite3.OperationalError: Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/dbus/service.py", line 711, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "src/hamster-service.py", line 269, in UpdateFactJSON
    return self.update_fact(fact_id, fact)
  File "/home/sh_zam/workspace/hamster/src/hamster/storage/storage.py", line 122, in update_fact
    self.__remove_fact(fact_id)
  File "/home/sh_zam/workspace/hamster/src/hamster/storage/db.py", line 747, in __remove_fact
    self.__remove_index([fact_id])
  File "/home/sh_zam/workspace/hamster/src/hamster/storage/db.py", line 844, in __remove_index
    self.execute("DELETE FROM fact_index where id in (%s)" % ids)
  File "/home/sh_zam/workspace/hamster/src/hamster/storage/db.py", line 936, in execute
    cur.execute(state, param)
sqlite3.OperationalError: near "(": syntax error
```

**Comment by ederag** [comment]
Thanks for the details.
Can't reproduce here, but that might be due to a different python/sqlite version (3.6.5 here).
Need to install 3.8 to check, although others seemed to run 3.8 fine.
What is you exact python version ?

There are a lot of other calls using this syntax (searched for `in (%s)`), so this is scary.
Or could it be that the `ids` string becomes weird in your case ?

**Comment by sh-zam** [comment]
Python version: `3.8.1`

I don't find anything weird with `ids`, debug log just before the crash:
```
hamster.storage.db - INFO - removing fact #dbus.Int32(1425) from index
hamster.storage.db - DEBUG - DELETE FROM fact_index where id in (dbus.Int32(1425)) ()
```

**Comment by ederag** [comment]
Looks like #477 then. 
Is your version lower than dbus-python-1.2.14 ?

**Comment by sh-zam** [comment]
Yep, that's it! Updating dbus-python fixed it.

**Comment by ederag** [comment]
Thanks for the confirmation.
Sorry you had to lose time on this, the python-dbus guys were really quick to fix this,
but apparently some distributions did not upgrade yet. What is your distribution ?

I'm interested in your fix, because it both fixes the issue and uses a list instead of building a string,
so looks cleaner. But where did you find this syntax ?
in the documentation, I found only [executemany](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.executemany).

**Comment by sh-zam** [comment]
I am using Manjaro.

> But where did you find this syntax ?

https://github.com/projecthamster/hamster/blob/fc8275268c73c102fe9b228f5be8baff9593e1ea/src/hamster/storage/db.py#L930-L932

^ in this I saw that, it is creating lists out of strings. So, I passed a list.

Now, I noticed that `ids` isn't a list, so the patch is wrong. I think, it working is purely coincidental! 

The correct way to call would be:

```self.execute(["DELETE FROM fact_index where id in (?)"], [(ids,)])```

Which I think would only confuse people, don't you think?

**Comment by ederag** [comment]
Agreed, that would be confusing.
In the standard case where `statement` is not a list, 
the statement and parameters are both encapsulated in a list, so that `zip` can be used:
https://github.com/projecthamster/hamster/blob/fc8275268c73c102fe9b228f5be8baff9593e1ea/src/hamster/storage/db.py#L934

We should probably use [executemany](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.executemany):
https://github.com/projecthamster/hamster/blob/fc8275268c73c102fe9b228f5be8baff9593e1ea/src/hamster/storage/db.py#L943

Current state is not good, 
but it does not seem wise to make this change so close to v3.0 though. Let me think.

**Comment by ederag** [comment]
Managed to reproduce with python3.8, indeed.
`executemany` is not better:
```python
  File "/home/ederag/share/prog/python/hamster/src/hamster/storage/db.py", line 843, in __remove_index
    self.executemany("DELETE FROM fact_index where id = (?)", ids)
  File "/home/ederag/share/prog/python/hamster/src/hamster/storage/db.py", line 947, in executemany
    cur.executemany(statement, params)
ValueError: parameters are of unsupported type
```
It wants a list of strings, so the ids would have to be converted to strings anyway.
That would make the code a bit less readable, in many places.

=> Check for version incompatibility.
The error will be clearer, and we do not touch sensitive code at this moment.
Is it OK ?

sh-zam:
>The error will be clearer, and we do not touch sensitive code at this moment.
Is it OK ?

Yep, seems like a good plan.

You pushed the commit [`fdb0e5`](https://github.com/projecthamster/hamster/commit/fdb0e5a3139fff20f788bd7a4d3e99237076b7e7) to this branch, you should push it to master as well :-)?

ederag:
Thanks, I forgot to change branches before fdb0e5a3139fff20f788bd7a4d3e99237076b7e7.
Fixed, but it turns out that reverted commits do not count in the credits, although they are valuable.
Still learning... I should have added a Co-Authored-By line to the merge commit.
At least it is your PR that is cited in the merge commit.
Thanks again for your input !

sh-zam:
>  Still learning... I should have added a Co-Authored-By line to the merge commit.

It's okay :smile:
Thank you maintaining the software :pray:

PR #539, opened 2020-02-05
Add backward compatibility of timestamps of facts-table

lutra:
In older versions of the database one can find milliseconds in
timestamps of facts-table. When loading, one gets a ValueError exception.
Fix this by adding a default microsecond.

### DIFF
--- a/src/hamster/storage/db.py
+++ b/src/hamster/storage/db.py
@@ -1006,7 +1006,7 @@ class Storage(storage.Storage):
 
 # datetime/sql conversions
 
-DATETIME_LOCAL_FMT = "%Y-%m-%d %H:%M:%S"
+DATETIME_LOCAL_FMT = "%Y-%m-%d %H:%M:%S.%f"
 
 
 def adapt_datetime(t):
@@ -1019,7 +1019,11 @@ def convert_datetime(s):
 
     s is in bytes
     """
-    return dt.datetime.strptime(s.decode('utf-8'), DATETIME_LOCAL_FMT)
+    date_string = s.decode('utf-8')
+    if '.' not in date_string:
+        date_string = date_string + '.0'
+
+    return dt.datetime.strptime(date_string, DATETIME_LOCAL_FMT)
 
 
 sqlite.register_adapter(dt.datetime, adapt_datetime)
--- a/src/hamster/storage/db.py
+++ b/src/hamster/storage/db.py
@@ -1006,7 +1006,7 @@ class Storage(storage.Storage):
 
 # datetime/sql conversions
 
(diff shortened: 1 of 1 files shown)


ederag:
Thanks for the PR !
using `sqlitebrowser ~/.local/share/hamster/hamster.db &`,
my database has no milliseconds, back to november 2010, so your one must be even older ?

There were plans for using [`fromisoformat`](https://docs.python.org/3.8/library/datetime.html?highlight=datetime#datetime.date.fromisoformat) anyway.
But it is ahead of schedule; I'll have to backport that from python3.7 to support python3.6.

This is a sensitive part. Could you please copy-paste one of these timestamps ?
I'd like to test before merging.

ederag:
Your fix definitely worked, thanks !

Here is a slightly modified version, that truncates the string instead of appending,
because the time will be truncated by the hamster datetime anyway.
Gave up on the isoformat idea;
too much code, too little time to test; and there might be gotchas down the line.

lutra:
> my database has no milliseconds, back to november 2010, so your one must be even older ?

> This is a sensitive part. Could you please copy-paste one of these timestamps ?
> I'd like to test before merging.

Here are two entries from it:
```shell
sqlite> select id, start_time, end_time from facts;
1124|2020-02-04 08:28:39.988564|2020-02-04 08:51:52.898341
1125|2020-02-04 08:51:52.898341|2020-02-04 09:01:00
```
Beneith usage of the wonderful _hamster-cli_, I'm using the cinnamon [hamster-applet](https://github.com/linuxmint/cinnamon-spices-applets/commits/master/hamster@projecthamster.wordpress.com). Maybe they write timestamps with microseconds to dbus? Had a look but did not find anything.

Anyway, thank you.

ederag:
Thanks for the additional information.
[Here](https://github.com/linuxmint/cinnamon-spices-applets/blob/ce72523215992eb6136da539048de0a465a3dead/hamster%40projecthamster.wordpress.com/files/hamster%40projecthamster.wordpress.com/applet.js#L563-L565) seems where they `AddFact` through dbus.
Is it still the case with current master ? What is your python version ?

lutra:
Yes I have the same version. When updating cinnamon-applets you get the latest versions of the master branch of the [repository](https://github.com/linuxmint/cinnamon-spices-applets).

The python version I start hamster with is 3.7.6.

ederag:
I meant the version of hamster itself.
With python3.7.3 the timestamps are `2020-02-07 08:43:00`, as they should.
How do you start hamster exactly ?
If started from the command line, please give the output of the `version` command.
But it really looks like you are running an older version. 
Finally, have you [killed the hamster daemons](https://github.com/projecthamster/hamster#kill-hamster-daemons) ?

Issue #506, opened 2020-01-07
Checking for program 'glib-genmarshal'   : not found


aasutossh:
After installing dependencies, when I run ./waf configure build I get the following error.
```
./waf configure build
Setting top to                           : /home/aasutossh/git/hamster 
Setting out to                           : /home/aasutossh/git/hamster/build 
Checking for program 'glib-genmarshal'   : not found
```
Tried searching for the issue, no lead.
Thank you.

ederag:
That's a new one.
*What is your system ?*
In general, please specify the distribution and the version of hamster you are using.
You may try to type `glib-genmarshal` in a terminal, 
that usually gives instructions on how to install the missing program.
Please paste back these instructions here, so others can benefit.

aasutossh:
So sorry.
I am using pop os 19.10.
I used git clone to get the hamster(must be latest version).

Writing `glib-genmarshal` in the terminal outputs the following:
```
Command 'glib-genmarshal' not found, but can be installed with:

sudo apt install libglib2.0-dev-bin
```
Installing this worked for me.
Build successful.
Thank you so much.

ederag:
Thanks for the success feedback.

Note that from `apt info libglib2.0-dev`
>  This package is needed to compile programs against libglib2.0-0. It contains
 development utilities typically run during compilation and should not be
 installed directly. Use libglib2.0-dev instead.

So you might want to issue `sudo apt install libglib2.0-dev` to be more standard.

aasutossh:
Well, snap way seems much easier.
I went for snap way.
Thanks.

PR #546, opened 2020-02-10
Add libglib2.0-dev-bin Ubuntu dependency

Forage:
At least Ubuntu 19.10 needs it for glib-genmarshal to be available

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ https://snapcraft.io/hamster-snap
 ###### Ubuntu (tested in 19.04 and 18.04)
 
 ```bash
-sudo apt install gettext intltool python3-gi-cairo python3-distutils python3-dbus python3-xdg libglib2.0-dev-bin
+sudo apt install gettext intltool python3-gi-cairo python3-distutils python3-dbus python3-xdg libglib2.0-dev
 # and for documentation
 sudo apt install gnome-doc-utils yelp
 ```
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ https://snapcraft.io/hamster-snap
 ###### Ubuntu (tested in 19.04 and 18.04)
 
 ```bash
-sudo apt install gettext intltool python3-gi-cairo python3-distutils python3-dbus python3-xdg
+sudo apt install gettext intltool python3-gi-cairo python3-distutils python3-dbus python3-xdg libglib2.0-dev-bin
 # and for documentation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,7 +63,7 @@ https://snapcraft.io/hamster-snap
 ###### Ubuntu (tested in 19.04 and 18.04)
 
 ```bash
-sudo apt install gettext intltool python3-gi-cairo python3-distutils python3-dbus python3-xdg
+sudo apt install gettext intltool python3-gi-cairo python3-distutils python3-dbus python3-xdg libglib2.0-dev-bin
```
ederag:
```suggestion
sudo apt install gettext intltool python3-gi-cairo python3-distutils python3-dbus python3-xdg libglib2.0-dev
```
From `apt info libglib2.0-dev-bin`:
>  This package is needed to compile programs against libglib2.0-0. It contains
 development utilities typically run during compilation and should not be
 installed directly. Use libglib2.0-dev instead.

ederag:
Thanks ! That's very helpful, as it was [in my todo list](https://github.com/projecthamster/hamster/projects/1#card-31310276) to check whether #506 was isolated or not.
That makes at least two of you, do definitely worth adding.
Please check the suggestion though.
Strange thing is that's not required in any of the `lubuntu` virtualboxes tried so far. Nevermind.

README.md
Forage: Installing libglib2.0-dev instead of just libglib2.0-dev-bin causes a butt load of other dependency packages being installed without a need for them. Which is a bit of a shame.

It makes me wonder how much of a need we have for glib-genmarshal to begin with. I suppose we can't just drop the requirement?
If not than yes, better to stick to the recommendation to prevent issues for the user later on.

ederag: > It makes me wonder how much of a need we have for glib-genmarshal to begin with. I suppose we can't just drop the requirement?

Good questions. If anyone finds evidence that's it's not required 
and ways to silence the complaint, 
that would be great, maybe after v3.0 as time is short.

> If not than yes, better to stick to the recommendation to prevent issues for the user later on.

Agreed, thanks again !

ederag: Looks like I cannot push this recommendation myself
(there was a checkbox to allow edits by maintainer to allow that).
It's OK, but it's up to you now.

Forage:
Change applied

Issue #433, opened 2019-08-24
Future of hamster v3.x?


GeraldJansen:
This issue is intended for discussion of the future of the upcoming maintenance versions (v3.0, v3.1, etc.) of the "legacy" hamster time tracker.

IIUC, the main purpose of this effort is to get the gtk3+python3 version of legacy hamster back into shape as a reliable tool which can be easily installed as a system application on recent versions of mainstream linux distributions. Presumably v3.0 would be a candidate for package managers to replace the old hamster-applet packages which have been dropped from recent distributions (cf. #356, #399, #421). 

Thanks to the formidable efforts of @ederag, it seems to me that we are in pretty good shape w.r.t. these objectives. Perhaps adequate testing and documentation of changes relative to hamster-applet remain as issues, but the main question is whether there are any package managers still interested in providing and maintaining hamster packages. A Snap package for hamster (#418) could well be a good alternative.

In the meantime, the rewrite as `hamster-lib/cli/gtk/dbus` is still advertised as being the main thrust of future development, in spite of the fact that it seems to be completely stalled for about 2-3 years now. To put it bluntly, it looks lifeless (ie. dead). 

Personally, before putting further effort into the project, I would like to know where things are heading. 

If there is any hope for reviving the rewrite, my preference would be to try to join that effort (with some modest contributions).  Otherwise the v3.x line should be re-branded as the main thrust and some issues/enhancements currently assigned to hamster-gtk should be reassigned to v3.1, v3.2 etc. and ... we should drive on. Maybe some of the work from the rewrite could be considered for back-porting. I would be willing to continue with occasional contributions, but as I said, it would be really nice to know where the project is heading (if anywhere).

ederag:
There is an interest for a debian package (#369), and @mwilck is packaging for openSUSE.

It would be good to hear from the rewrite devs.

The rewrite is still a good idea, especially as a base for new features,
but as time is going by the legacy code is becoming familiar,
and its maintainability increases. 
It could be the stable version for the years to come,
even if the rewrite goes forward again.

hedayat:
Well, as there is no manpower for it, I agree that the rewrite effort is dead. And, even if it revives again, it can (and actually does!) have a different name (hamster-cli/gtk/...?).

[...]

PR #549, opened 2020-02-11
Restore edit, remove dialogs

### LINKED ISSUE #548: "Add earlier activity" dialog no longer shows up from Shell extension
With the 3.0-beta, clicking on the "Add earlier activity" menu entry from the GNOME Shell extension no longer makes the dialog pop up. Other entries like the overview and settings work as expected.

An intermediate build on 30 October worked as expected.
Calling the same dialog from the overview dialog works as well.

Is there any way I can have a Shell extension provide me with log messages?
Could this be another case of some outdated/duplicate files lingering somewhere due to file/folder name changes?

ederag:
WIP: for now it work only with the 0 id (just add),
so only "Add earlier activity" will work (fix #548).

### DIFF
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -101,7 +101,7 @@ class Hamster(gtk.Application):
     Actions should eventually be accessible via Gio.DBusActionGroup
     with the 'org.gnome.Hamster.GUI' id.
     but that is still experimental, the actions API is subject to change.
-    Discussion with "external" developpers welcome !
+    Discussion with "external" developers welcome !
     The separate dbus org.gnome.Hamster.WindowServer
     is still the stable recommended way to show windows for now.
     """
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -31,6 +31,7 @@ gi.require_version('Gtk', '3.0')
 from gi.repository import GLib as glib
 from gi.repository import Gtk as gtk
 from gi.repository import Gio as gio
+from gi.repository import GLib as glib
 
 import hamster
 
@@ -145,6 +146,10 @@ class Hamster(gtk.Application):
 
     def on_startup(self, data=None):
         logger.debug("startup")
+        # Must be the same as application_id. Won't be required with gtk4.
+        glib.set_prgname(self.get_application_id())
+        # localized name, but let's keep it simple.
+        glib.set_application_name("Hamster")
 
     def _open_window(self, name, data=None):
         logger.debug("opening '{}'".format(name))
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -123,8 +123,8 @@ class Hamster(gtk.Application):
     def add_actions(self):
         # most actions have no parameters
         # for type "i", use Variant.new_int32() and .get_int32() to pack/unpack
-        for name in ("about", "add", "edit", "overview", "preferences"):
-            data_type = glib.VariantType("i") if name == "edit" else None
+        for name in ("about", "add", "clone", "edit", "overview", "preferences"):
+            data_type = glib.VariantType("i") if name in ("edit", "clone") else None
             action = gio.SimpleAction.new(name, data_type)
             action.connect("activate", self.on_activate_window)
             self.add_action(action)
@@ -162,15 +162,15 @@ class Hamster(gtk.Application):
                 self.about_controller = About(parent=_dummy)
                 logger.debug("new About")
             controller = self.about_controller
-        elif name == "add":
-            if not self.fact_controller:
-                self.fact_controller = CustomFactController(parent=self)
-                logger.debug("new CustomFactController")
-            controller = self.fact_controller
-        elif name == "edit":
-            if not self.fact_controller:
-                id_ = data.get_int32()
-                self.fact_controller = CustomFactController(parent=self, fact_id=id_)
+        elif name in ("add", "clone", "edit"):
+            if self.fact_controller:
+                # Something is already going on, with other arguments, present it.
+                # Or should we just discard the forgotten one ?
+                logger.warning("Fact controller already active. Please close first.")
+            else:
+                fact_id = data.get_int32() if data else None
+                self.fact_controller = CustomFactController(name, fact_id=fact_id,
+                                                            parent=self)
                 logger.debug("new CustomFactController")
             controller = self.fact_controller
         elif name == "overview":
@@ -191,6 +191,20 @@ class Hamster(gtk.Application):
         controller.present()
         logger.debug("window presented")
 
+    def present_fact_controller(self, action, fact_id=0):
+        """Present the fact controller window to add, clone or edit a fact.
+
+        Args:
+            action (str): "add", "clone" or "edit"
+        """
+        assert action in ("add", "clone", "edit")
+        if action in ("clone", "edit"):
+            action_data = glib.Variant.new_int32(int(fact_id))
+        else:
+            action_data = None
+        # always open dialogs through actions,
+        # both for consistency, and to reduce the paths to test.
+        app.activate_action(action, action_data)
 
 class HamsterCli(object):
     """Command line interface."""
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -27,8 +27,10 @@ import argparse
 import re
 
 import gi
-gi.require_version('Gtk', '3.0')
+gi.require_version('Gdk', '3.0')  # noqa: E402
+gi.require_version('Gtk', '3.0')  # noqa: E402
 from gi.repository import GLib as glib
+from gi.repository import Gdk as gdk
 from gi.repository import Gtk as gtk
 from gi.repository import Gio as gio
 from gi.repository import GLib as glib
@@ -169,8 +171,7 @@ class Hamster(gtk.Application):
                 logger.warning("Fact controller already active. Please close first.")
             else:
                 fact_id = data.get_int32() if data else None
-                self.fact_controller = CustomFactController(name, fact_id=fact_id,
-                                                            parent=self)
+                self.fact_controller = CustomFactController(name, fact_id=fact_id)
                 logger.debug("new CustomFactController")
             controller = self.fact_controller
         elif name == "overview":
@@ -188,6 +189,18 @@ class Hamster(gtk.Application):
         if window not in self.get_windows():
             self.add_window(window)
             logger.debug("window added")
+
+        # Essential for positioning on wayland.
+        # This should also select the correct window type if unset yet.
+        # https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html
+        if name != "overview" and self.overview_controller:
+            window.set_transient_for(self.overview_controller.window)
+            # so the dialog appears on top of the transient-for:
+            window.set_type_hint(gdk.WindowTypeHint.DIALOG)
+        else:
+            # toplevel
+            window.set_transient_for(None)
+
         controller.present()
         logger.debug("window presented")
 
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -98,7 +98,12 @@ def fact_dict(fact_data, with_date):
 class Hamster(gtk.Application):
     """Hamster gui.
 
-    Can be accessed across D-Bus with the 'org.gnome.Hamster.GUI' id.
+    Actions should eventually be accessible via Gio.DBusActionGroup
+    with the 'org.gnome.Hamster.GUI' id.
+    but that is still experimental, the actions API is subject to change.
+    Discussion with "external" developpers welcome !
+    The separate dbus org.gnome.Hamster.WindowServer
+    is still the stable recommended way to show windows for now.
     """
 
     def __init__(self):
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -496,7 +496,9 @@ Example usage:
                 assert len(args.action_args) == 1, (
                        "edit requires exactly one argument, got {}"
                        .format(args.action_args))
-                action_data = glib.Variant.new_int32(int(args.action_args[0]))
+                id_ = int(args.action_args[0])
+                assert id_ > 0, "received non-positive id : {}".format(id_)
+                action_data = glib.Variant.new_int32(id_)
             else:
                 action_data = None
             app.activate_action(action, action_data)
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -28,6 +28,7 @@ import re
 
 import gi
 gi.require_version('Gtk', '3.0')
+from gi.repository import GLib as glib
 from gi.repository import Gtk as gtk
 from gi.repository import Gio as gio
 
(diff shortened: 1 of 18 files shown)


src/hamster-windows-service.py
```diff
@@ -41,6 +41,14 @@ def _open_window(self, name):
         subprocess.run("hamster {} &".format(name),
                        shell=True)
 
+    @dbus.service.method("org.gnome.Hamster.WindowServer")
+    def edit(self, id=None):
+        if id:
+            # Need an edit <id> command in cli
+            raise NotImplementedError()
```
mwilck: Sorry this won't do it. The Shell extension also uses edit with <id> (that's the all-important button I'm using all the time to fix my wrong start times ...)

If you want to maintain compatibility, you need to revert 0cffe9c2c7e669a3b0557a15f61e2a8ea538be21.

mwilck:
From the comment in `hamster-windows-service.py`

> new code should proabably access the GtkApp actions directly

Could you give an example how to do that? Also, there's only an "add" action, no "edit" action...

src/hamster-windows-service.py
ederag: That's not helping, on the contrary.
Please read carefully the links before posting: https://github.com/projecthamster/hamster/issues/548#issuecomment-584676139

ederag:
@mwilck please read carefully https://github.com/projecthamster/hamster/pull/549#issue-373701974

mwilck:
> @mwilck please read carefully [#549 (comment)](https://github.com/projecthamster/hamster/pull/549#issue-373701974)

I saw it but it's not helpful. The "edit" button is vastly more important than "Add Earlier activity" (for my use case at least). Why don't you simply revert 0cffe9c2c7e669a3b0557a15f61e2a8ea538be21?  I can confirm that reverting that commit fixes both issues ("Add Earlier Activity" and "Edit button"), whereas your patch deliberately does not.

Or do you seriously want me to create a new issue for this?

ederag:
@mwilck Again it is clear that you do not fully understand the message I want to convey in my comments.
The comment I linked says "for now", and "WIP" means "Work In Progress", 
meaning I'm on it, not satisfied yet, so working on the "edit" thing [if I was not distracted, that is].
So all your comments in this review/thread were just noise for me.
On the contrary your [test and confirmation in PR 551](https://github.com/projecthamster/hamster/pull/553#commitcomment-37249369) was appreciated, for instance.

mwilck:
@ederag, I know what "WIP" means. I fail to understand why pointing out the solution to an issue counts as "distraction" to you. There'd be no need to invest any additional work into this issue if you just reverted the offending commit.

ederag:
> There'd be no need to invest any additional work into this issue if you just reverted the offending commit.

And add technical debt ? Ugh.
That's exactly the kind of statement that adds friction.
Please never *tell* me what to do, as we clearly do not have the same goals for hamster,
and we'll make more progress.

mwilck:
> Please never _tell_ me what to do, 

I meant to make a suggestion, nothing more. Sorry if I sounded rude.

> as we clearly do not have the same goals for hamster,

We have not? My current goal is to help with the beta test and bug squashing, to be able to to push an "official" hamster 3.0 package to openSUSE. And I'd like to be able to continue using the Shell extension, including the "edit activity" button. Does that conflict with your goals?

ederag:
Technical debt reduction seems more important to me, that's the main difference I see.

Still WIP, as the fact controller can only be opened once
(because despite the name it's not a `Controller` yet)
Not an issue from the command line.
Does it work good enough for the extension, should it be merged without waiting further ?

mwilck:
OK - if I just understood what you mean with "technical debt"... I understand that you would like to get rid of the legacy interface. I'm not against changing the extension to use the new API, but it'll take more time. First I'd need to understand how to call the `org.gnome.Hamster.GUI` interface.

mwilck:
Your code doesn't work in development mode, because the shell command `hamster` will start the installed version of hamster, not the development one. I just sent https://github.com/ederag/hamster/pull/2 as a fix.

mwilck:
So yeah, I guess this would work, too. But I'm not happy about this change at all.

     def _open_window(self, name):
         subprocess.run("hamster {} &".format(name), shell=True)

You had sophisticated DBus message passing in clean python, and now, instead, you fork/exec first a shell, and then fork/exec another hamster instance. You think this is an improvement wrt the code we had before? What about the risks of invoking commands built from user input with a shell? I could pass in 

     GLib.Variant("s", "0 & tar cf /tmp/leaked.tar ~/.ssh")

... with the expected effect ... at least you need to check whether the "id" parameter is actually an integer. I'd strongly suggest you don't use `shell=true`.

I don't like saying it because I much appreciate the work you do, but I think you're on the wrong track here. What's wrong with just calling `dialogs.edit.show()`, like the old code did? Why is that "technical debt"?

ederag:
@mwilck 
> OK - if I just understood what you mean with "technical debt"... I understand that you would like to get rid of the legacy interface.

I use the standard meaning of "[technical debt](https://en.wikipedia.org/wiki/Technical_debt)".
Let's consider the alternative 
1) fix a bug right now with a hack or removing an improvement to the code, or
2) fix a bug later with a cleaner code.

From your past recommendations, it *seemed* that you tend to favor the quick fix 
much more often than me. Is that a false impression ?

We have communication some issues, but I'm pretty sure that can be overcome with time.
Let me explain the intention about the legacy interface in #433.
Hopefully you'll see that the extension can stay as it is for a long time.

mwilck:
In my opinion, the "clean" solution is to modify the Shell extension, and possibly other clients, to use the new API (which would need to be extended by an "edit" Action). Neither my simple revert and this `fork()/exec()` approach would classify as clean, future-proof solutions to me.

In general, I think we both know that a compromise must be reached between the ideal ambition and the limited amount of resources. As you maybe realized, I've devoted more resources than usual to hamster this week. I won't be able to keep this up, next week I'm into system programming and storage again. Perhaps you feel relieved about that :grin:  

Back to the point, this approach here is acceptable for me, although I'd still prefer the revert. You just need to fix that shell parameter passing issue.

ederag:
@mwilck Messages crossed (no notifications).

> Perhaps you feel relieved about that

A bit relieved surely, as I will not be able to take many other days off to follow the pace;
your input is sometimes useful although unusually time consuming for me.
But I'm glad that you helped improving the quality of hamster :smile:

> In my opinion, the "clean" solution is to modify the Shell extension, and possibly other clients, to use the new API (which would need to be extended by an "edit" Action)

Exactly. That's the plan.
But the plan is also to provide `hamster-windows-service.py` as a backward compatibility layer,
as long as possible, if needed by clients.

Hopefully improvements to the new `org.gnome.Hamster.GUI` will drive these clients to
use the actions directly.


> Your code doesn't work in development mode

Sure, as it was only a compatibility layer for installed hamster.
But your proposition is interesting, thanks !

> I could pass in `GLib.Variant("s", "0 & tar cf /tmp/leaked.tar ~/.ssh")`

From a quick glance, the variant is packed/unpacked by `new_int32()`/`get_int32()`, 
that would just error. Need to check though.

Before the new "edit", all actions had no arguments so it was safe.
So it looked like a good temporary measure,
before learning how to do that cleanly
(already done a good deal of research, some things are unclear yet).

Right now it will have to do, with adequate security measures if needed.

> What's wrong with just calling dialogs.edit.show(), like the old code did? Why is that "technical debt"

That used to be the recommended way, about ten years ago.
The new standard way should be much cleaner, *in the end*.

mwilck:
> > In my opinion, the "clean" solution is to modify the Shell extension, and possibly other clients, to use the new API (which would need to be extended by an "edit" Action)
> 
> Exactly. That's the plan.
> But the plan is also to provide `hamster-windows-service.py` as a backward compatibility layer,
> as long as possible, if needed by clients.

The clean way to do the latter would be to set up `hamster-windows-service` as a "DBus proxy" that simply translates the legacy DBus calls into the new ones, talking to the `org.gnome.Hamster.GUI` API, Shouldn't be that hard, actually, except that we have no "edit" action, as mentioned before.

ederag:
> The clean way to do the latter would be to set up hamster-windows-service as a "DBus proxy" that simply translates the legacy DBus calls into the new ones, talking to the org.gnome.Hamster.GUI API,

No, that's most probably not the clean way. There's a new gnome way.
I searched for some time, good leads, clearly possible, possibly dead easy *once done*,
but no clear-cut example for our case. We'll see after v3.0.

> except that we have no "edit" action

376f1b0073e83af1f5e3f611bb9b471716a59ee4 ? That's the purpose of this PR.

> Shouldn't be that hard

Beware of assumptions, especially when judging my choices. 
Not saying they are always good, there are mistakes, 
but they are all the result of a thorough and continuous process of maintenance.

Are you sure to have the necessary overview ?
Answering some of your posts drains a lot of my time (actually more than coding in this PR !),
because I do not want other users to be scared too much for nothing.
It's clear that you want the best for this project; so maybe it would be better to 
focus on testing (that's the most important thing to do actually) as you have already done
and possibly research such as https://github.com/projecthamster/hamster/pull/556#issue-374294456,
and let me decide with a bit more slack.

mwilck:
> There's a new gnome way.
> I searched for some time, good leads, clearly possible, possibly dead easy _once done_,
> but no clear-cut example for our case. We'll see after v3.0.

I'd like to understand what that new GNOME way is.

> > except that we have no "edit" action
> 
> [376f1b0](https://github.com/projecthamster/hamster/commit/376f1b0073e83af1f5e3f611bb9b471716a59ee4) ? That's the purpose of this PR.

I wasn't talking about the CLI. What I meant was an "edit" action in the `org.gtk.Actions` API. But OK, that API isn't supposed to be used, at least not he way I thought. Understood. Beware that others might think the same though - normally the purpose if a DBus API is to be called.

> Beware of assumptions, especially when judging my choices.

The old `hamster-windows-service`, which you called "techical debt", had a lean and powerful design. It ran in the background, ready to show helpful UI for entering data into hamster, without fork() and exec(). It was nicely suited for other applications to be used as DBus service. It isn't obvious to me what superior replacement you plan to design. Your allusions about more modern and cleaner design don't clarify much. What I see in the current code is a step backward.

You say I should trust your choices. So that's what I'll do. I'll have another look when your new code is implemented.

ederag:
> What I meant was an "edit" action in the org.gtk.Actions API

It's there, first one:
Screenshot_20200214_184223

> What I see in the current code is a step backward.

Agreed. Except that's "retract only to jump higher".

ederag:
Ready for early testing.
Not satisfied with the window positioning, still trying.

ederag:
Window positioning is now correct (dialogs on top of their transient-for),
except in gnome/wayland where the compositer places them where it wants to.

`dialogs` and `Dialog` have been removed (c1bf53b342acdc1f64c794e3f8e0b7ecb10fc3d5)
because they might be already broken and can not be tested any longer. 
Their maintenance would be a burden, let's use the major release opportunity.

Is anyone else using them anyway ? 
We could make a wrapper.

ederag:
This morning changes were purely cosmetics, *no* need to retest !
Just fixed the last merge commit message, it was pointing to #2 instead of ederag#2,
and fixed the imports order.

GeraldJansen:
Sorry to report that it seems slightly broken again (v3.0-beta-39-gb10c9bb2). From xfce-hamster-plugin, clicking 'Add Earlier Activity' opens the Overview window instead of the Add window. If an overview window has already been opened from the command line I get this error message:
```
[hamster]$ pkill -ef hamster.*service && hamster overview
python3 killed (pid 7714)
python3 killed (pid 7719)
[1]+  Done                    hamster overview
ERROR:dbus.connection:Unable to set arguments (None,) according to signature 'i': <class 'TypeError'>: an integer is required (got type NoneType)
  File "/usr/bin/hamster", line 144, in on_activate_window
  File "/usr/bin/hamster", line 174, in _open_window
    self.fact_controller = CustomFactController(name, fact_id=fact_id)
  File "/usr/lib/python3/dist-packages/hamster/edit_activity.py", line 94, in __init__
    self.fact = runtime.storage.get_fact(fact_id)
  File "/usr/lib/python3/dist-packages/hamster/client.py", line 190, in get_fact
    return from_dbus_fact_json(self.conn.GetFactJSON(id))
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 147, in __call__
    **keywords)
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 643, in call_blocking
    message.append(signature=signature, *args)
TypeError: an integer is required (got type NoneType)
```

AFAICT, all other functionality seems to be okay.

ederag:
Thanks for the report.
I just realized you need a `pkill -ef hamster` to kill also any `hamster` (the `Gtk.Application`) running,
otherwise the `hamster` script will find the old application 
and connect to it instead of launching a new one.
That feels bold (there might other processes with 'hamster' in their command).
We need a `hamster quit` command to quit the GUI.

Looking into the actual issue now.

GeraldJansen:
The same behaviour can be triggered from d-feet by executing the edit() method of the WindowServer with either no input or 0 as the fact id.

ederag:
Thanks, that helped a lot !
It would be tempting to handle `edit 0` in the actions as well,
but now we have the new `add` action for that.
Better stick to one way only IMO.

GeraldJansen:
Hmm, with 4e593e6 xfce-hamster-plugin now does nothing for `Add Earlier Activity` because it uses edit() with an empty argument list ([here](https://github.com/projecthamster/xfce4-hamster-plugin/blob/f9f5e2177fb7791f90575a1c381dfafb92ce7e51/panel-plugin/view.c#L120-L146)).

ederag:
Well, I misunderstood the [dbus doc](https://dbus.freedesktop.org/doc/dbus-python/) then
> dbus-python might not be the best D-Bus binding for you to use. dbus-python does not follow the principle of “In the face of ambiguity, refuse the temptation to guess”

and in the [tutorial](https://dbus.freedesktop.org/doc/dbus-python/tutorial.html#data-types)
> Unlike Python, D-Bus is statically typed - each method has a certain signature representing the types of its arguments, and will not accept arguments of other types.

It seems to work now.
Tested by leaving the `method input` empty in d-feet, 
as `GLib.Variant("()", ())` only crashes d-feet flat.

GeraldJansen:
Everything seems to be working fine from the xfce plugin now. Thanks.

ederag:
Phew :slightly_smiling_face:
Thanks for the feedback and the clear pointers !

mwilck:
@ederag, thanks for your continued work and persistence!

ederag:
@mwilck Thanks, so that seems to work with the gnome-shell-extension as well, good.

src/hamster-cli.py
```diff
@@ -95,7 +98,12 @@ def fact_dict(fact_data, with_date):
 class Hamster(gtk.Application):
     """Hamster gui.
 
-    Can be accessed across D-Bus with the 'org.gnome.Hamster.GUI' id.
+    Actions should eventually be accessible via Gio.DBusActionGroup
+    with the 'org.gnome.Hamster.GUI' id.
+    but that is still experimental, the actions API is subject to change.
+    Discussion with "external" developpers welcome !
```
ederag:
```suggestion
    Discussion with "external" developers welcome !
```

ederag:
Screenshot_20200218_121732
This PR is the "WIP: ..." line: 3 unique visitors (us three, thanks for testing !).

PR #554, opened 2020-02-12
Fix completion in categories TreeView

sh-zam:
Because `key` was being returned in lowercase, matching would fail.


ederag:
Thanks, looks good !
Found out the issue it solves:
You need a former upper case category (say "CAT")
There was no way to autocomplete this one, even by typing "CAT".

PR #556, opened 2020-02-12
hamster: explicitly set application ID and prgname on startup

### LINKED ISSUE #541: Icon missing / App tracking broken in Gnome/Wayland
I think #538 broke app tracking somehow.

I just updated and restarted Hamster, now GNOME Shell doesn't manage to figure out that the running instance corresponds ot the desktop file.

Here's what I see in the dash, observe how the running icon is separate from the launcher:

dash-hamster

And here's alt+tab:

alt-tab-hamster

mwilck:
Under wayland, passing application_id to Gtk.Application.Init()
is not sufficient to set the application ID and program name
so that the shell sees it. See e.g.

https://www.programcreek.com/python/example/87806/gi.repository.GLib.set_prgname
https://gitlab.gnome.org/GNOME/gtk/issues/653
https://gitlab.gnome.org/GNOME/gtk/commit/72ec8963d7df0e39f688838e42445c869e4875c9

This should fix #541.

### DIFF
--- a/src/hamster-cli.py
+++ b/src/hamster-cli.py
@@ -98,13 +98,11 @@ class Hamster(gtk.Application):
     Can be accessed across D-Bus with the 'org.gnome.Hamster.GUI' id.
     """
 
-    __app_id__ = "org.gnome.Hamster.GUI"
-
     def __init__(self):
         # inactivity_timeout: How long (ms) the service should stay alive
         #                     after all windows have been closed.
         gtk.Application.__init__(self,
-                                 application_id=self.__app_id__,
+                                 application_id="org.gnome.Hamster.GUI",
                                  #inactivity_timeout=10000,
                                  register_session=True)
 
@@ -144,8 +142,10 @@ class Hamster(gtk.Application):
 
     def on_startup(self, data=None):
         logger.debug("startup")
-        glib.set_application_name(self.__app_id__)
-        glib.set_prgname(self.__app_id__)
+        # Must be the same as application_id. Won't be required with gtk4.
+        glib.set_prgname(self.get_application_id())
+        # localized name, but let's keep it simple.
+        glib.set_application_name("Hamster")
 
     def _open_window(self, name, data=None):
         logger.debug("opening '{}'".format(name))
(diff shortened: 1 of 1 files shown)


src/hamster-cli.py
```diff
@@ -141,6 +144,8 @@ def on_activate_quit(self, data=None):
 
     def on_startup(self, data=None):
         logger.debug("startup")
+        glib.set_application_name(self.__app_id__)
+        glib.set_prgname(self.__app_id__)
```
bochecha: Are both of those necessary?

mwilck: No, not to fix the icon name. But I think this way we're on the safe side.

ederag:
Interesting, and thanks for the links ! Need to test and read.

ederag:
That yields both under wayland and gnome the left image.
Screenshot_20200213_070607 Screenshot_20200213_092815

which is exactly what `hamster --name=org.gnome.Hamster.GUI` gave;
and that I found ugly at first (too much information for a casual user).
Right image is the result of c3ae7b1e2c86e924f549fc334ebb7c31089e637b

Your [link to commit](https://gitlab.gnome.org/GNOME/gtk/commit/72ec8963d7df0e39f688838e42445c869e4875c9) eventually lead me to
https://developer.gnome.org/gtk4/stable/gtk-migrating-3-to-4.html#id-1.6.4.3.13
>  In GTK 4 we want the application's GApplication 'application-id' (and therefore the D-Bus name), 
the desktop file basename and Wayland's xdg-shell app_id to match. 
In order to achieve this with GTK 3.x call g_set_prgname() 
with the same application ID you passed to GtkApplication. 
Rename your desktop files to match the application ID if needed.

> The call to g_set_prgname() can be removed once you fully migrated to GTK 4.

So the `set_application_name` is redundant with `app_id=`, let's remove it,
and just keep `set_prgname`.

Let's just use the [`application_id`](http://lazka.github.io/pgi-docs/Gio-2.0/classes/Application.html#Gio.Application.get_application_id) property,
that clarifies the changes to be made then (a single line).

mwilck:
Wow, thanks for sorting this out. Not really obvious for the casual programmer, is it?

mwilck:
Fine with me, please merge at your discretion.

ederag:
@mwilck, @bochecha Thanks for the confirmations (I take the thumb up as one).
And indeed, it's not obvious; nice team work !

Issue #552, opened 2020-02-12
Schema problem when running uninstalled


GeraldJansen:
Running in development/uninstalled mode fails because of uninstalled schema.
Using current master (21568b04):
```
[src]$ python3 hamster-cli.py 
(hamster-cli.py:6999): GLib-GIO-ERROR **: 09:20:18.546: Settings schema 'org.gnome.Hamster' is not installed
Trace/breakpoint trap (core dumped)
```

mwilck:
I don't think you can do much about that. I believe the schema has to be installed in a system location. But I may be wrong.

AFAICS you wouldn't have this problem if you installed once (from any version) and used development mode after that.

ederag:
Indeed, just hoped to get an answer to my [question](https://stackoverflow.com/q/59567042/3565696), then forgot.

Workaround:
```bash
glib-compile-schemas --targetdir=build/data data
GSETTINGS_SCHEMA_DIR=build/data python3 src/hamster-service.py &
GSETTINGS_SCHEMA_DIR=build/data python3 src/hamster-cli.py
```
With your clear issue, it should be easy to find for devs.
Let's keep it open until we manage to automate that.

PR #557, opened 2020-02-12
Update README with comment concerning Development

GeraldJansen:
See #552.


ederag:
Agreed, this is well placed.

PR #558, opened 2020-02-12
A couple appdata changes

bochecha:
This renames the appdata file so it follows the same naming convention as the desktop file, the icons, the DBus service, …

In addition, this provides the old `hamster.desktop` name, so that software centers like GNOME Software or KDE Discover can know that `hamster.desktop` and `org.gnome.Hamster.GUI.desktop` are really the same application. This is useful in a few cases:

*   if your LTS distro provides an older version in its repository with the `hamster.desktop` name Flathub (soon?) provides a newer version with the name `org.gnome.Hamster.GUI.desktop`, users would be very confused to see 2 different entries called "Hamster" with the same icon

    with this change, GNOME Software would show a single app, but would allow users to choose from which source to install it

*   if users had submitted reviews in the ODRS for the older `hamster.desktop`, then e.g GNOME Software can know to display them for the newer `org.gnome.Hamster.GUI.desktop`


ederag:
@bochecha Looks good, and thanks for the clean explanations.
I suppose that this time there was no clear reference to link to ?
ORDS certainly is https://odrs.gnome.org/

@extraymond This change would break snap, as it gets informations from the metadata ([line 4](https://github.com/extraymond/hamster-snap/blob/dcaecfaf2e88e9f9d421f7e095305ff6f97df3c4/snap/snapcraft.yaml#L4)).
but from https://snapcraft.io/hamster-snap the stable is from 2020-01-30 so it should stay unaffected.
The edge is probably broken already (https://github.com/extraymond/hamster-snap/pull/5). So there's no reason to delay.
Sorry for the inconvenience. https://github.com/extraymond/hamster-snap/pull/6 should fix this.

bochecha:
Ah, my apologies for breaking the Snap, I missed that.

>  I suppose that this time there was no clear reference to link to ?

I'm honestly not sure. The best I can think of is this doc, but it is specific to Flatpak: https://docs.flatpak.org/en/latest/conventions.html#appdata-files

The convention about using reverse DNS to name everything has been latent for about a decade, but not particularly well followed outside of GNOME.

Flatpak actually makes that convention mandatory: it's how it associates all those resources together to the app, how it knows which DBus service/icon/launcher an app can export to the rest of the OS, etc…

ederag:
@bochecha Thanks. No need to apologize at all, on the contrary,
these name changes will be [needed anyway for gtk4](https://developer.gnome.org/gtk4/stable/gtk-migrating-3-to-4.html#id-1.6.4.3.13) and 
being standard will surely help in other areas.

PR #559, opened 2020-02-13
Desktop and i18n work

mwilck:
This started out with me realizing that we had no translations in our `.desktop` file. I thought that's not ideal for a desktop program. I tried to figure out how to generate them and stumbled upon the recent move from intltool to gettext, and figured that the `.po` files haven't been updated for quite a while. It was a pretty rough ride because I'm not familiar with either **waf** or **gettext**.

Eventually I have come up with a way to rebuild the `.po` files from the current source, and _optionally_ commit them in in git. (That's the big "update translations" commit. I've reviewed the `de.po` file and the results seemed reasonable to me). I also added some missing German translations.

Please have a look.


ederag:
Please hold on,
I have plenty of documentation for i18n, and plans.
This is work in progress.
We'll discuss that *after* v3.0.

Please do not even try to PR the first commit separately, it's good practice to keep the Exec line.
I'll find the reference if you wish.

mwilck:
Pushed again with the removal of the `Exec=` lines reverted.

ederag:
Some commits look promising, 
but that can not be merged right now, we'll discuss after v3.0,
thanks for your understanding !

matthijskooijman:
> This started out with me realizing that we had no translations in our .desktop file.

Hm, while looking for something else, I noticed #497 earlier today, which seems to remove translations of the desktop file (without really mentioning this in the commit messages, so I'm not sure why). Maybe you @mwilck already know this, but otherwise it might be useful background (haven't looked at this PR at all yet, though).

mwilck:
I noticed that the translations were gone, and I made this PR to add them back. I found it strange to create a 3.0 release with a desktop file without any translations.

I'm definitely not an i18n expert. Things have changed a lot in this area, you find notes everywhere that tools that were previously used are now deprecated and out of support. I fiddled around with this for a day or two until the results looked ok-ish to me. Obviously I couldn't review all translations. After @ederag's last comment (Feb 18), I didn't bother any more.

@ederag has made some changes to the waf code after this PR that prevent merging it directly. I guess it's fixable with some mechanic changes (variable renames), but it's been 3 months, I've forgotten most of the stuff I did here.

mwilck:
... and no, I don't have slightest idea why these translations had been removed.

PR #561, opened 2020-02-14
build: Add a Flatpak manifest

bochecha:
This allows building the master branch of Hamster with a simple command:

$ flatpak-builder _build build-aux/flatpak/org.gnome.Hamster.json

---

For those who don't know Flatpak much, here are a few additional instructions.

1. you need to install Flatpak, Flatpak Builder and configure the Flathub remote: https://flatpak.org/setup/
2. you will need to install the GNOME platform and sdk to build;
3. to make rebuilds faster you want to use the `--ccache` and `--force-clean` options;
4. once the app is built, you need to export it to a local repo, configure that local repo, then install from that repo; this can be automated with the `--install` and `--user` options;

All in all, the following two commands should do the right thing:

```
$ flatpak install flathub org.gnome.Platform//3.34 org.gnome.Sdk//3.34
$ flatpak-builder --force-clean --ccache --user --install _build build-aux/flatpak/org.gnome.Hamster.GUI.json
```

If this is merged, we can look at having automatic builds in your CI, published to a development repo, etc…

And then I can help you send the next release to Flathub if you're interested. slightly_smiling_face

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ hamster-time-tracker-*.tar.gz
 .lock-waf*
 build
 *.deb
+.flatpak-builder
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "build-aux/flatpak/shared-modules"]
+	path = build-aux/flatpak/shared-modules
+	url = https://github.com/flathub/shared-modules
--- a/build-aux/flatpak/org.gnome.Hamster.GUI.json
+++ b/build-aux/flatpak/org.gnome.Hamster.GUI.json
@@ -1,0 +1,64 @@
+{
+  "id": "org.gnome.Hamster.GUI",
+  "runtime": "org.gnome.Platform",
+  "runtime-version": "3.34",
+  "sdk": "org.gnome.Sdk",
+  "command": "hamster",
+  "finish-args": [
+    "--socket=x11",
+    "--socket=wayland",
+    "--own-name=org.gnome.Hamster"
+  ],
+  "modules": [
+    "shared-modules/intltool/intltool-0.51.json",
+    {
+      "name": "dbus-python",
+      "buildsystem": "simple",
+      "build-commands": [
+        "pip3 install --no-deps --no-use-pep517 --prefix=${FLATPAK_DEST} ."
+      ],
+      "cleanup": [
+        "*.la"
+      ],
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://files.pythonhosted.org/packages/source/d/dbus-python/dbus-python-1.2.16.tar.gz",
+          "sha256": "11238f1d86c995d8aed2e22f04a1e3779f0d70e587caffeab4857f3c662ed5a4"
+        }
+      ]
+    },
+    {
+      "name": "pyxdg",
+      "buildsystem": "simple",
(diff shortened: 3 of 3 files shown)


ederag:
Thanks for the proposition,
same answer as for any package, here is the one about [snap](https://github.com/projecthamster/hamster/pull/496#issuecomment-568523325),
I'd prefer to split responsibility, as least for now. 
(I'm fully [supporting](https://github.com/extraymond/hamster-snap/pull/5), following the repo, just not taking responsibility)
Is it possible to do that in an external repo for flatpak as well ?

bochecha:
It certainly is, yes. In fact that's how it works for Flathub.

If I may though, there are quite a few benefits to having this in the upstream repository:

1. people don't need to figure out what options to pass to Waf, look at the build failures to try and find what dependencies they are missing, because everything is fully described in the manifest;

2. people can just open the Hamster source code in GNOME Builder and start hacking, the "build" button will build Hamster within Flatpak, the "run" button runs it in a sandbox, which leads to a pretty great developer experience and makes [onboarding newcomers much easier](https://wiki.gnome.org/Newcomers/BuildProject#Download_and_Build_a_Project);

3. that can even become an integral part of your workflow, where the CI builds a Flatpak bundle for each merge request and exports it somewhere, making it trivial to test proposed changes; GNOME does that for lots of its apps now, it has made everybody's life much easier, especially the non-developers (e.g designers or translators) who now don't need to build apps and maintain development environments;

I'll understand if you still don't want it in. I do think it would be a mistake, but it's entirely your choice to make and I'll respect your decision.

If you decide not to take this in, then I won't maintain a Flatpak build of the master branch myself, I'll simply wait for the release and push it to Flathub then. :slightly_smiling_face:

ederag:
OK, that's seducing.

Point 1 is probably not related to being in the master repo itself.
Snap is external, and users do not have to care about waf either.

Yet, points 2 and 3 are a bit different.
Flatpak being so close to gnome, and hamster being a Gtk app, 
there are additional advantages with respect to other package systems.
And I can see that development could be improved by having all in one place.
That could deserve a special policy.

The v3.0 coming release (+ main job) is a bit overwhelming,
it would not be reasonable for me to learn a new framework now;
could that be discussed again just after the release ?

Or maybe I could assign automatically any flatpak related issue to you,
and merge blindly any PR involving only those files ?

bochecha:
> The v3.0 coming release (+ main job) is a bit overwhelming,
> it would not be reasonable for me to learn a new framework now;
> could that be discussed again just after the release ?

Oh of course.

> Or maybe I could assign automatically any flatpak related issue to you,
> and merge blindly any PR involving only those files ?

I'd rather not be the only one understanding this stuff. Teach a person to fish, and all that.

There's no rush here. :slightly_smiling_face:

ederag:
> I'd rather not be the only one understanding this stuff. Teach a person to fish, and all that.

I'm always curious, and progressively learned some of snap for instance,
so you would not be alone.
But likewise, until things settle down, I would like the *responsibility* of flatpak to be on other shoulders.
Thanks for your comprehension, so we'll discuss again about that after v3.0.

ederag:

ederag:

matthijskooijman:
I'm closing this, since #610 has been merged instead.

Issue #491, opened 2019-12-11
Allow bypass (0) and override to None (-1) start/end in DBus AddFact


ederag:
#490 seemed to fix the shell-extension (https://github.com/projecthamster/hamster-shell-extension/pull/316#issuecomment-564145295),
but that overrides the parsed fact start_time with `hamster_now ` as a default.

Here is the 2012 state:
https://github.com/projecthamster/hamster/blob/80110affd6f0c3a91d739c09a78dffe7c06976a1/src/hamster-service#L117-L121
https://github.com/projecthamster/hamster/blob/80110affd6f0c3a91d739c09a78dffe7c06976a1/src/hamster/lib/stuff.py#L322-L323
so start and end times were taken from the parse string, and overridden only 
if the separately given `start_time` and `end_time` were non-zero.

This prevents overriding the end_time with None for instance.

##### Proposition

Use `0` to mean "do not mess around with the parsed fact"
(just ignore this argument)
Use `-1` as a special value to mean "replace with `None`"

This way the full power of the new parser (including datetimes and relative times parsing)
can be exploited in the shell-extension as well.

For instance one might try this branch:
https://github.com/ederag/hamster/tree/AddFact-parse-times

jmberg:
Hmm. So the gnome shell extension doesn't include any times in the string at all, as far as I can tell. Would that still use "hamster_now()" then

Of course the user could enter "10:00 something" but that's not really suggested in the shell extension UI - other places pre-fill the current time there and put the caret after it to indicate that you can change this around, but the shell extension just has a single entry field and even labels this "What are you doing?", and auto-completes only the fact name, not any times.

(Technically, you could even argue that one should be able to enter a fact like "10:00 meeting" where the "10:00" is part of the fact name, not a time. But neither does that really matter, nor do I have any idea how that would've behaved previously. I assume it would've parsed it already for a long time.)

AddFact-parse-times branch looks like it should work, but I can only test it later.

mwilck:
Agreed with @jmberg, the typical usage of the extension is to enter only an activity name or activity@category. With the current extension and hamster 2.2.2, it's actually *possible* to enter a start time via the extension's text input field, but not an end time (it isn't parsed correctly, at least). Also, as @jmberg noted, there's no autocompletion if a start time is used in the input field. I have never used this, I always enter just the current activity, and if I need to fix the start time, I click the "edit" button to start the hamster UI and apply changes there. Admittedly not the perfect workflow, but I've got used to it.

[...]

PR #562, opened 2020-02-14
No start override

### LINKED ISSUE #560: start time ignored from xfce-hamster-plugin
When entering a verbatim activity such as `10:15 work@home` from the xfce-hamster-plugin, the specified starting time is ignored and instead the current time is used. The same happens using a relative time such as `-20`.

@mwilck Would you mind checking if this works with the gnome shell extension?

Note: testing hamster version v3.0-beta-16-gc78079f8+ (i.e. with 0cffe9c temporarily reverted).
Note: to get the plugin working, I had to `./configure --prefix=/usr --with-icon=hamster`.

ederag:
This is #491 again,  supplemented with "if the parsed fact has no start, use now".
Should be better than #492, 

fix #560 while keeping full backward compatibility with existing clients.

@jmberg the `10:00 meeting` activity can be entered either by tricking
(e.g. prepending a character: `.10:00 meeting`),
or (that's new) by using double dashes `--` as explicit missing times (see help).
Here That would be `-- -- 10:00 meeting`.

 @mwilck Thanks for your test case in #491. Now the time would not be ignored,
so in some sense #492 was not, while this one is fully backward compatible.
If a client wants "activity without range" (*not recommended at all*), 
it could just prepend `-- -- ` to the string:
Screenshot_20200214_155923

Screenshot_20200214_155643

### DIFF
--- a/src/hamster-service.py
+++ b/src/hamster-service.py
@@ -135,7 +135,7 @@ class Storage(db.Storage, dbus.service.Object):
 
     # facts
     @dbus.service.method("org.gnome.Hamster", in_signature='siib', out_signature='i')
-    def AddFact(self, fact_str, start_time, end_time, temporary=False):
+    def AddFact(self, fact_str, start_time, end_time, temporary):
         """Add fact specified by a string.
 
         To fully use the hamster fact parser, as on the cmdline,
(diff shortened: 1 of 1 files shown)


ederag:
0cffe9c2c7e669a3b0557a15f61e2a8ea538be21 reverted to allow testing only :wink:

mwilck:
Thanks, this works better than before.

GeraldJansen:
Tested this also with xfce-hamster-plugin and "everything" (well, everything I tested) seems to be working fine now.

ederag:
Thanks !
Removed the revert, prior to merging.

PR #565, opened 2020-02-16
Packagers skip options

### LINKED ISSUE #550: Minor packaging issues (waf)
I've been working on the openSUSE package for 3.0-beta, and found that two "features" of the new waf code are rather non-helpful for packagers:

 * automatic updating of the icon cache
 * automatic compilation of gsettings schemas

For packaged hamster, both need to be done at **installation time**, not at built time. Distros usually have macros or scripts in place to do this automatically. I can see that doing these steps is useful for people who run the manual installation procedure. For packagers, these steps create useless files that need to be removed again before creating the actual package. Perhaps we could find a way to disable these steps for distro builds?

If not, no worries, I've quickly patched this out and can continue to do so.

ederag:
At the instigation of @mwilck (thanks for the clear and kind request !), 
packagers can now use two new "skip" options (fix #550)
```
./waf --help | grep -i1 packagers
  --prefix=PREFIX       installation prefix [default: /usr]
  --skip-gsettings      skip gsettings schemas build and installation (for packagers)
  --skip-icon-cache-update
                        skip icon cache update (for packagers)
```

That escalated a bit in a long-wanted overhaul of the wscripts.
`wscript-build` was changed to a fully valid python script named `wscript`
(it is clearer that the `build` method in that script will be called by the root wscript,
 instead of the code being mysteriously "included", with `bld` not explicitly defined locally)


mwilck:
This looks good to me, except that I find the argument renaming to intrusive at this "shortly before RC" stage.

ederag:
> Calling the argument bld or conf depending on stage seems to be common practice for waf.

That's an historical cruft. And a one that got me puzzled a lot when discovering this code.
By following the [waf book style](https://waf.io/book/#_build_related_commands), the entry will be smoother for new coders.
`ctx` stands for *the* context, exposed to functions, not unrelated objects.

> argument renaming to intrusive at this "shortly before RC" stage.

Well, that might have been dangerous, 
but I took time to check thoroughly the search&replace outcome (kdevelop/kate is very good at that), and test,
and decided to include it *now* precisely because packagers won't be surprised
if their patches break upon a major version change.

mwilck:
Ok, fine with me.

ederag:
@mwilck Thanks for testing.
Packagers, please do file an issue if something else can be done to ease the packaging process.

matthijskooijman:
I was playing with these options a bit and was a bit surprised that these must be given during *build*, rather than during *install* when the commands are run. It might be more sensible to check for these options during *install*, though especially for the gsettings, I'm not entirely sure how to approach this. Also, it's not a very big deal, of course.

ederag:
The options must be set at the `configure` step, to store them.
Here is the rationale:
https://github.com/projecthamster/hamster/blob/ff1ab0e8addc19f7d3b84af2077947ae8fd8df61/wscript#L58-L62
Maybe that should be made more visible by separating `configure` and `build` in the instructions.

Issue #570, opened 2020-02-21
Access to secondary databases


ederag:
There is a need to access other hamster databases as well, not only the main `hamster.db`,
e.g. for merging a foreign database into the main one (#340).
As recently [discussed](https://github.com/GeraldJansen/hamster-tool/issues/1#issuecomment-589394913), this is not possible using the `org.gnome.Hamster` D-Bus service,
which is tied to the main database.

One possibility:
1) add an option (`--no-dbus` ?) to [hamster-cli](https://github.com/projecthamster/hamster/blob/7138d1aff96dbd0d15410bfd26284904885e672e/src/hamster-cli.py) to bypass D-Bus and use [db.Storage](https://github.com/projecthamster/hamster/blob/7138d1aff96dbd0d15410bfd26284904885e672e/src/hamster/storage/db.py#L48) directly.
2) add an option (`--db-file=<filename>`) to [hamster-cli](https://github.com/projecthamster/hamster/blob/7138d1aff96dbd0d15410bfd26284904885e672e/src/hamster-cli.py).
2) Of course this access is risky
(there is no central server to avoid simultaneous write access in this case), 
so if the server is running, it should be forbidden to access the server database.
(that one will remain hard-coded)

Still thinking, open to discussion.

GeraldJansen:
It seems to me that you underestimate sqlite3's capabilities for handling concurrent requests from multiple processes and overestimate the risk associated with accessing hamster.db directly, by-passing the hamster-service D-Bus interface (eg. [here](https://github.com/projecthamster/hamster/issues/434#issuecomment-589220846)). From a bit of reading, for example
- https://stackoverflow.com/questions/14217249/sqlite-and-concurrency and
- https://sqlite.org/lockingv3.html,

in my assessment, considering the small size our hamster.db files, the processing speed of even oldish laptops/desktops and the low likelihood of concurrent write requests, the "risk" associated with accessing hamster.db without using the D-Bus interface is extremely low.

ederag:
Thanks for the links ! Lots of food for thought.

ederag:
The links were spot-on,
and reading carefully is fully convincing that sqlite is robust and perfect for hamster,
sql writes from different processes just queue up. 
(timeout is possible, but the database integrity remains)
What a relief, thanks Gerald !

Next step is to check that the `start_transaction` / `end_transaction` blocs are enough
to protect from database changes during `python` code execution,
in particular this call in the middle of `__add_fact__` should be checked:
[...]

Issue #340, opened 2016-11-08
Merging hamster.db files


unfa:
I happen to have a few hamster.db files that I want to merge.
How can I do this?

I'd love to be able to import old hamster.db files into Hamster having them merged into the current one.

elbenfreund:
Each ``hamster.db`` is a independend ``sqlite3`` database. I doubt there is a solid way to achieve what you ask for without significant import / conflict resolution logic (which does not exist as far as I know).
You could probably try turning each file into a csv and then use the dbus interface to ``add`` each row as a fact to your master ``hamster.db`` hoping that hamsters "squeeze" mechanism solves any conflicts.

Project hamster and its various sub-components is currently undergoing some major changes. We prepare the introduction of a rewritten codebase for most of the underlying functionality. A direct consequence of this is that it is unlikely that any open/new bugs within the current/old codebase will be fixed (unless someone steps up and offers to do so) as most resources currently available will be invested in making the rewrite prime time ready.

unfa:
Ok, thank you.

 have two .db files that should have no conflicts, I don't know the exact structure of the databases, but there are no duplicated events between the two files.

I'd like to just read all events listed and write them to a 3rd file.

I don't know if they need to be sorted in some way. Do you know any tools that could help me manipulate the .db files? Is there a graphical SQLite database editor?

neekfenwick:
@unfa did you make any progress on this? I'm moving from a laptop where I run hamster to a desktop where I used to run hamster, and want to merge the db on each.  I imagine some kind of INSERT using the resultset of a SELECT would be possible in theory, did you try this and what was your general outcome? (disaster, success, partial mess, etc)

jamesfryer:
I have used this script. Export records as a TSV then run this script on the file. Finally run the resulting file (using source command) on the target machine.

```
#! /bin/bash
# Take a time tracker TSV export and convert to series of hamster-cli start commands

if [[ "$1" == "" ]]; then echo "Please give me filename of Hamster TSV file"; exit 1; fi

filename="$1"
awk -F"\t" 'NR>1 {print "hamster-cli start \"" $1 "@" $5 "," $6 "\" \"" $2"-"$3"\""}' "$filename"
```

GeraldJansen:
[...]

Issue #222, opened 2015-02-27
Mac OS X port


iliv:
Hello,

I'm a big fan of Project Hamster. It is hands down the best time tracking software there is out there. Design and usability are unmatched by any other. It is most intuitive to use application in its class I've ever run into. It makes a pretty much dull and tedious task of time tracking an interesting exercise because I can feel how this little app alleviates most boring input efforts with memorized entries and tags. I could go on and on, statistic module albeit not extremely flexible and detailed is very decent and, well, looks great! A point that gets overlooked by many in software development world.

Anyway, I couldn't find a good enough replacement for Project Hamster on OS X -- despite the myriad of commercial applications -- nothing is quite as good. I've been checking out your wordpress site and github account in hopes to see OS X version for years but alas this is apparently not something you're planning to do.

I just wanted to contact you and let you know that there is a real need for a good application like project hamster on OS X. There's a gap to fill. Much like GnuCash OS X port is a kick-ass personal finance application on OS X that has no rivals even of commercial breed, had the Project Hamster been ported to OS X, it would've been the king of time tracking.

I know how open source software development is and I totally understand that OS X is not everyone's priority but I do wonder what it would take to port Hamster to OS X. Especially seeing how GnuCash is actively supported. There must be framework already established that could be taken advantage of and save Project Hamster developers a ton of time. 

Is there any "political will" at all? How demanding an effort like that would be? Heck, I so want to see Hamster on OS X that I'm considering to run a donation campaign or something.

Ivan

nchachereau:
Hi there! Thank you for all this enthusiasm, I am sure everyone who tries to devote some free time to hamster is glad you like it so much.

However, as we are only a few people, without very much time and doing this on the side... porting the app to a platform we do not even use is really not an option. At all :disappointed:  Especially since the current code base has so many bugs... So in the interest of cleaning the list of issue, I am marking this as wontfix and closing.

I hope you will find a way to run hamster on OS X (does anyone know whether this is possible? after all, GTK+ is more or less cross platform...) or another great application for your new platform of choice. Kind regards!

jeff-dagenais:
sad.

Any updates on this? Any other suggestions? Does the rewrite have a better chance of working on macOS?

ederag:
This would require someone with knowledge of macOS and dbus.
And yes, the rewrite uses up to date python packaging 
that has better chance to work cross-platform.

mapkyca:
I'm also trying to scratch this itch. 

[...]

PR #573, opened 2020-03-01
WIP: No dbus

ederag:
First part of #570, new option
```
  --no-dbus             (experimental) Direct access to the database. Another
                                process trying to write at the same time is untested yet.
```
It could clearly have been done more concisely, but hey, there it is :slightly_smiling_face: 

The only breaking change is passing `storage` as an argument instead of a centralized `runtime`.
But AFAIK no-one is using controllers or widgets. 
*Please chime in quickly if this is an overlook.*

`runtime` will stay there deprecated for a while to ensure compatibilty.

All changes that could have affected known companions 
have received backward compatibility measures. *Please chime in quickly if there was a leftover*

Still WIP, not well tested, for information.


mwilck:
I don't like this idea. dbus serves as a layer of separation/abstraction between the database and the clients using. Unless I'm very mistaken, that's a basic design principle of hamster. 

IMO the rationale in the PR is not strong enough to justify such a massive architectural change. If anything, we should rather follow @elbenfreund's idea to separate out `hamster-lib` *below* the dbus layer. Then access to the db would be possible by calling `hamster-lib` directly.

ederag:
That was part of the ongoing plan, as I said,
> It could clearly have been done more concisely, but hey, there it is slightly_smiling_face

I know you disagree with the plan. 
It has already been clarified that we have very different philosophies.
Let's not waste our energy, just close, and be on your own.

> dbus serves as a layer of separation/abstraction between the database and the clients using it.

Sure, that stays the default.
This PR is allowing another route, that can help in special cases.
e.g. https://github.com/projecthamster/hamster/issues/222#issuecomment-601931145 ends with
> Thanks for the help and the --no-dbus flag :)

GeraldJansen:
@mwilck Unlike you, I think having a --no-dbus option is a great idea.
- it permits isolation of dbus problems during development and testing 
- it frees you from the hassle of killing the dbus services continually during development when you are not working on dbus related issues
- it opens the door to be able to access more than one database through the storage layer, which would allow integration of database merging (eg. desktop and laptop)
- it could be useful in cases of dbus problems with snaps or flatpaks or on other operating systems (osx #222 -yay, Windows? - yuk)

Hamster itself works fine without dbus. (In fact I'm working slowly on my own hamster-lite fork in which I've ripped out the dbus support altogether). With hamster brought up by a desktop hotkey and closed by the ESC key, hamster usage is super efficient and out-of-your-face. 

Not all Hamster users are gnome-hamster-extension centric!

mwilck:
@GeraldJansen, dbus is in no way tied to the GNOME extension. I believe you are aware of that.

The usage of dbus is motivated by the idea of separating and abstracting tasks. By using DBus, end-user applications don't have to worry about database access, transactions, synchronization, concurrency etc., because there's only one application (the DBus service) that accesses the data base. It's actually beneficial for development, because data base access and UI functionality are separated and can be developed and tested separately. 

With this separation, the DBus layer allows easy development of "foreign" UI clients. The GNOME extension is just one example. IIRC you said you use the XFCE extension, which uses the same functionality. With the invention of the "hamster windows service", the original hamster authors went one step further and made it possible to create a new UI without implementing every dialog. You can implement the basic UI, and hand over to the original hamster for details. That was *true genius* and made the invention of the GNOME extension and other helpers feasible in the first place.

By ditching this concept, your "hamster-lite fork" becomes a very different type of application. That doesn't mean "inferior" or "worse", but quite far away from the original spirit in which hamster has been designed, as I understand it.

> * it permits isolation of dbus problems during development and testing
> * it frees you from the hassle of killing the dbus services continually

You and @ederag seem to see DBus mainly as a source of complication and bugs. I concur that it adds complexity if you want to pull up new UI features quickly. Rather than simply hacking a new field in the data base and adding an UI control for it, you have to consider how to transport the new piece of information through the available DBus methods, or possibly extend the DBus API, which is additional work and requires a lot of careful consideration. I still believe that it's good to have this separation, and that spending the additional effort would pay off in the long run. What you are referring to in the two bullets above actually IMO shows that we have *too little* separation between frontend and backend in the current code, not *too much*.

> * it opens the door to be able to access more than one database through the storage layer, which would allow integration of database merging (eg. desktop and laptop)

In which way is this related to DBus? Doing it needs a change to the storage layer in the first place. DBus could actually be helpful here, as the layers on top of the DBus protocol could be agnostic of how many database files the storage layer accesses, and where they are stored, and let the storage layer worry about it alone. 

I'm not sure what you mean with "database merging" - do you want to merge the storage in a one-shot operation, or do you want to able to use both databases concurrently and sort of merge them transparently / on the fly? In the former case, I suppose a dedicated command line tool `hamster-merge-db` would be the way to go. The biggest issue here, as with every merge operation, would be how to deal with conflicting entries. But DBus has nothing to do with that.

> * it could be useful in cases of dbus problems with snaps or flatpaks or on other operating systems

Every modern desktop has some sort of messaging system that would be usable as a separation layer. It doesn't have to be DBus, of course. I'm not opposed to generalizing the concept, if someone wants to tackle that.

GeraldJansen:
As far as I know, Gnome-hamster-extension is the only really "alive" project that relies on the hamster dbus services, aside from Hamster's own CLI. (Btw, I do not use the xfce plugin, I just did some testing of it as a contributor to this project.) 

This PR is about adding an option to make it possible to use Hamster, including the GUI and  CLI, without DBus, not about removing DBUS from Hamster. Unlike me, @ederag has always been a very strong supporter of maintaining a stable DBus interface.

The dbus services should not be a layer between the main UI and the DB backend(s). In fact, in contrast to what you have stated earlier, it seems to me that @elbenfreund designed hamster_cli and hamster_gtk to be based directly on hamster_lib, and he added hamster_dbus later to support external service consumers like gnome-hamster-extension. 

For some background on the issue of accessing more than one DB and merging DBs, see #340, #570,  and the external tool I developed [htool](https://github.com/GeraldJansen/hamster-tool/.) Imho, it would be better to have the tool integrated in the Hamster CLI (for better visibility and maintainance), and I had to do some convincing to bring @ederag around to agree.

I don't wish to argue endlessly, but it would be nice if you could be a bit more generous in your consideration of other's opinions.

mwilck:
> This PR is about adding an option to make it possible to use Hamster, including the GUI and CLI, without DBus, not about removing DBUS from Hamster.

I understand. My concern is that, once that door is open, people will slip through it rather than considering the more stony DBus path, and the DBus API will bit rot over time.

>  it seems to me that @elbenfreund designed hamster_cli and hamster_gtk to be based directly on hamster_lib, and he added hamster_dbus later to support external service consumers

Interesting, thanks for pointing this out. I suppose that warrants a discussion about the general architecture we aim at.

> I don't wish to argue endlessly, but it would be nice if you could be a bit more generous in your consideration of other's opinions.

I have stated my opinion, and tried to explain my reasons. That doesn't mean I don't respect other people's opinions. You are free to try and convince me; your argument about `hamster-lib` was a good start. Anyway, someone else will have to review the massive changes this PR contains (+813 -655 lines).

matthijskooijman:
(I typed most of this before the last two comments were posted, so I might need to separately respond to some of that, but no time right now).

Let me add my view on this as well.

As for the feature itself: Making hamster run without dbus, accessing the db storage directly, I can see merit in there. I do not think it should the only mode supported, or the default, but it I can see the merit for e.g. systems without dbus and local development and testing.

To properly support both, we would need a proper storage abstraction, to make sure that most code does not actually have to know whether it talks to dbus or the database directly. When I started writing this comment, I expected this would require significant refactoring to achieve, but now I looked at some of the commits in this PR (in particular https://github.com/projecthamster/hamster/pull/573/commits/f8b09e0db7c2068065e69f2f3b267b723c19e4b3 that seems to be the core of this commit), we actually have most, if not all, of this abstraction in place already with `hamster.storage.db` and `hamster.dbus.client` both implementing a (what looks to be) compatible `Storage` class.

So maybe adding `--no-dbus` option could make sense, and even be feasible with minimal effort.

As for needing `--no-dbus` for merging databases - I don't quite think that would be needed. I added a comment (https://github.com/projecthamster/hamster/issues/570#issuecomment-630167158) to the relevant issue with some further motivation, seemed like a better place than here.

As for this particular PR, it is quite big. However, it does seem that most changes are very cleanly separated into commits, which helps review. Still, for a lot of commits it is not immediately clear to me what their purpose is exactly (i.e. why are they needed or helpful to work towards `--no-dbus`) and for some I suspect they are just unrelated improvements (which might be better placed in a separate PR). Also, it looks like the "add --no-dbus option" adds the option, but then a lot of subsequent commits do additional refactoring to make that option work (or maybe fix running without that option, I'm not sure). If so, I'd rather have these refactorings happen earlier, so the coder remains working after every commit.

Then, even with this PR improved for easier review, it is still a quite non-trivial change. Personally, I might not find enough time to properly review and test this to merge it (that is, I think there might be other (more pressing or more personally motivating) things I would rather spend my available Hamster time on. I suspect the same might hold for the rest of the maintainer team. Then, it seems [@ederag has also lost interest](https://github.com/projecthamster/hamster/issues/590#issuecomment-621068164) in further contributions (unless I'm misinterpreting that comment), so I don't see this PR moving forward anytime soon.

mwilck:
> To properly support both, we would need a proper storage abstraction, to make sure that most code does not actually have to know whether it talks to dbus or the database directly.

IMO this works only if we strictly enforce any additions to be made in a way that works through DBus, too; otherwise the DBus interface will fall behind.

The non-native clients mostly aren't written in Python. In other languages, the "storage abstraction" used is exactly the `org.gnome.Hamster` DBus interface. Plus we also have "UI abstraction" provided by `org.gnome.Hamster.WindowServer`.

matthijskooijman:
> IMO this works only if we strictly enforce any additions to be made in a way that works through DBus, too; otherwise the DBus interface will fall behind.

Yeah, as I said, I think the dbus interface should just remain the default. And both `Storage` implementations should remain compatible (which could be enforced by running some identical tests against both, maybe, but is also somewhat implied since (I think) the dbus service also uses the db `Storage` implementation, as its own backend, right?

> The non-native clients mostly aren't written in Python. In other languages, the "storage abstraction" used is exactly the org.gnome.Hamster DBus interface. Plus we also have "UI abstraction" provided by org.gnome.Hamster.WindowServer.

I think this change would only affect the main hamster app (cli and gui), not any of these non-native clients (which would of course not work in setups without dbus, but you can't have your cake *and* eat it, I guess).

MTindev:
Hello,

Thanks to this PR I'm able to run `hamster-cli` on macOS 12.6 (Monterey, on M1). Just for that, thanks @ederag !

PS : I do not know if there is a more recent version (commit / release) which brings this option ?

ederag:
**If** you are a **new** to hamster indeed
(since I do not have to care so much about others, I made a very good change that would upset existing users),
advances can be seen in the [branches](https://github.com/ederag/hamster/branches) page of my fork.

Development is super slow now. It's very good and stable for me (also contains #597 and more), but no-one else is testing it.
You are welcome to use it (it remains GPL of course).

PR #577, opened 2020-03-05
fix off-by-one bug for hamster export tsv

GeraldJansen:
This PR is a workaround for a bug concerning the range.end in some date ranges (#576).

A better fix would address the issue in hamster.lib.datetime.Range.from_start_end(), but my attempts there created other problems.

PR #580, opened 2020-03-08
Fix handling of datetime arguments to from_start_end

### LINKED ISSUE #576: Off-by-one-day in hamster export tsv
I recently upgraded to hamster 3.0.0, and now noticed `hamster export tsv` started to return slightly too much data. It seems there is an off-by-one-day error in the "to" date. For example:

```
matthijs@grubby:~$ hamster export tsv 2020-01-21 2020-01-22
activity        start time      end time        duration minutes        category        description     tags
Reizen  2020-01-21 08:55        2020-01-21 10:00        65.0    MKIT            Work
Reizen  2020-01-22 07:55        2020-01-22 08:26        31.0    MKIT            Work
Reizen  2020-01-23 18:09        2020-01-23 19:09        60.0    Day-to-Day
```

(I trimmed some extra activities for privacy and clarity)

This should, AFAIU (and it always did) return activities on the 21st and
22nd, but it also includes the 23rd. I originally noticed this with
february 29th and suspect an issue with leap years, but it seems to
happen with any date I've tried so far.

I have not investigated yet, first wanted to at least document this.
Maybe someone else can try to reproduce this?

ederag:
Here is how I'd fix #576.

### DIFF
--- a/src/hamster/lib/datetime.py
+++ b/src/hamster/lib/datetime.py
@@ -590,11 +590,10 @@ class Range():
         Convenient to ease backward compatibility,
         and to handle either hdays or datetimes.
         """
-
         if isinstance(start, Range):
             assert end is None, "end cannot be passed together with a Range"
             return cls(start.start, start.end)
-        elif isinstance(start, datetime):
+        elif (start is None) or isinstance(start, datetime):
             # This one must come first,
             # because inheritance order is datetime < date < pdt.date.
             pass
@@ -608,8 +607,13 @@ class Range():
         elif isinstance(start, pdt.date):
             # transition from legacy
             start = hday.from_pdt(start).start
+        else:
+            raise TypeError(
+                "\n    First argument should be either Range, None, datetime or hday;"
+                "\n    received {}".format(type(start))
+                )
 
-        if isinstance(end, datetime):
+        if (end is None) or isinstance(end, datetime):
             # same as above
             pass
         elif isinstance(end, hday):
@@ -617,6 +621,10 @@ class Range():
         elif isinstance(end, pdt.date):
             # transition from legacy
             end = hday.from_pdt(end).end
+        else:
+            raise TypeError(
+                "\n    Second argument should be either None, datetime or hday;"
+                "\n    received {}".format(type(start)))
 
         return cls(start, end)
 
--- a/src/hamster/lib/datetime.py
+++ b/src/hamster/lib/datetime.py
@@ -595,7 +595,7 @@ class Range():
             return cls(start.start, start.end)
         elif (start is None) or isinstance(start, datetime):
             # This one must come first,
-            # because inheritance order is datetime < date < pdt.date.
+            # because inheritance order is datetime < pdt.datetime < pdt.date.
             pass
         elif isinstance(start, hday):
             # Idem, beware of the inheritance order;
@@ -716,4 +716,3 @@ class timedelta(pdt.timedelta):
     def total_minutes(self):
         """Return the duration in minutes (float)."""
         return self.total_seconds() / 60
-
--- a/src/hamster/lib/datetime.py
+++ b/src/hamster/lib/datetime.py
@@ -590,27 +590,35 @@ class Range():
         Convenient to ease backward compatibility,
         and to handle either hdays or datetimes.
         """
+
         if isinstance(start, Range):
-            assert end is None, "range and end are mutually exclusive"
-            range = start
-        else:
-            if isinstance(start, hday):
-                day = start
-                start = day.start
-                if end is None:
-                    end = day.end
-            elif isinstance(start, pdt.date):
-                # transition from legacy
-                start = hday.from_pdt(start).start
+            assert end is None, "end cannot be passed together with a Range"
+            return cls(start.start, start.end)
+        elif isinstance(start, datetime):
+            # This one must come first,
+            # because inheritance order is datetime < date < pdt.date.
+            pass
+        elif isinstance(start, hday):
+            # Idem, beware of the inheritance order;
+            # hday < date < pdt.date.
+            day = start
+            start = day.start
+            if end is None:
+                end = day.end
+        elif isinstance(start, pdt.date):
+            # transition from legacy
+            start = hday.from_pdt(start).start
 
-            if isinstance(end, hday):
-                end = end.end
-            elif isinstance(end, pdt.date):
-                end = hday.from_pdt(end).end
+        if isinstance(end, datetime):
+            # same as above
+            pass
+        elif isinstance(end, hday):
+            end = end.end
+        elif isinstance(end, pdt.date):
+            # transition from legacy
+            end = hday.from_pdt(end).end
 
-            range = Range(start, end)
-
-        return range
+        return cls(start, end)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


GeraldJansen:
Hmm, tests/stuff_tests works fine, but hamster crashes:
```
  File "/usr/bin/hamster", line 149, in on_activate_window
  File "/usr/bin/hamster", line 184, in _open_window
    self.overview_controller = Overview()
  File "/usr/lib/python3/dist-packages/hamster/overview.py", line 477, in __init__
    self.find_facts()
  File "/usr/lib/python3/dist-packages/hamster/overview.py", line 531, in find_facts
    self.facts = self.storage.get_facts(start, end, search_terms=search)
  File "/usr/lib/python3/dist-packages/hamster/client.py", line 156, in get_facts
    for fact in self.conn.GetFactsJSON(dbus_range, search_terms)]
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 147, in __call__
    **keywords)
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 653, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Python.sqlite3.InterfaceError: Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/dbus/service.py", line 711, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/usr/libexec/hamster/hamster-service", line 335, in GetFactsJSON
    for fact in self.get_facts(range, search_terms=search_terms)]
  File "/usr/lib/python3/dist-packages/hamster/storage/storage.py", line 161, in get_facts
    return self.__get_facts(range, search_terms)
  File "/usr/lib/python3/dist-packages/hamster/storage/db.py", line 734, in __get_facts
    datetime_to))
  File "/usr/lib/python3/dist-packages/hamster/storage/db.py", line 906, in fetchall
    cur.execute(query, params)
sqlite3.InterfaceError: Error binding parameter 1 - probably unsupported type.
```

ederag:
Confirmed, forgot to restart the server :man_facepalming: . Time to get out, I'm no longer into it...
I'll try to fix that later though.

src/hamster/lib/datetime.py
```diff
@@ -590,25 +590,33 @@ def from_start_end(cls, start, end=None):
...
-
-            if isinstance(end, hday):
-                end = end.end
-            elif isinstance(end, pdt.date):
-                end = hday.from_pdt(end).end
-
-            range = Range(start, end)
+        elif isinstance(start, datetime):
+            # this one must come first,
+            # because isinstance(datetime(), date) is True
```
matthijskooijman: I was a bit confused by this comment, since it seems there is no `isinstance(datetime(), date)` below at all. Looking more closely, I see that `pdt` is actually the `datetime` import. Maybe this comment could be adapted too:

    # because isinstance(datetime(), pdt.date) is True

That would have made things more clear for me.

src/hamster/lib/datetime.py
```diff
@@ -590,25 +590,33 @@ def from_start_end(cls, start, end=None):
         Convenient to ease backward compatibility,
         and to handle either hdays or datetimes.
         """
+
         if isinstance(start, Range):
             assert end is None, "range and end are mutually exclusive"
             range = start
-        else:
```
matthijskooijman: You moved a lot of code out of this `else`, but is that really needed? Even more, doesn't this break passing a `Range` object? AFAICS, the line

    range = Range(start, end)

now runs even when a `Range` is passed, overwriting the `range = start` above?

ederag: Sure, I'm downright ashamed about this commit.
The first if should return.

It was a sunday evening quick fix, and it looks like I'm no longer into it.
I'll try to fix it this evening.

src/hamster/lib/datetime.py
ederag: `date` inherits from `pdt.date`.
I'll add a clarification though.

mwilck:
So we have this and #577, and both are closed ... ?

matthijskooijman:
Maybe @ederag plans to submit a fixed PR later? If so, note that you can *also* do a (force) push to the branch for an existing PR to automatically update the PR (this is often better than starting a new PR, since any comment history will be kept).

ederag:
> Maybe @ederag plans to submit a fixed PR later?

Indeed (with force-push). Closed to lower the publicity.
Please just give me some time.

ederag:
Got it (hopefully...). Just a few tests to do before pushing again.

ederag:
Indeed, to avoid mistakes, I would restrict to `hday` for new code.
`date` is there mainly to be inherited by `hday`,
and as a temporary measure, until `edit_activity.py` has been updated to use `hday`.

src/hamster/lib/datetime.py
```diff
@@ -591,26 +591,42 @@ def from_start_end(cls, start, end=None):
         and to handle either hdays or datetimes.
         """
         if isinstance(start, Range):
-            assert end is None, "range and end are mutually exclusive"
-            range = start
+            assert end is None, "end cannot be passed together with a Range"
+            return cls(start.start, start.end)
+        elif (start is None) or isinstance(start, datetime):
+            # This one must come first,
+            # because inheritance order is datetime < date < pdt.date.
```
matthijskooijman: As pointed out by @mwilck, I think this comment should be:
```suggestion
            # because inheritance order is datetime < pdt.datetime.
```

Code would not need to be changed.

matthijskooijman:
> Nit wrt the comment: AFAICS, datetime inherits only pdt.datetime, not date.

Good point, I added a suggestion for that.

> can start be an instance of pdt.datetime() ? If yes, where is that case handled?

I think this is not a supported case. The old API was to pass `pdt.date` (or maybe `date`), the new API is `hday`, but it `datetime` might also be used when passing data through `from_start_end` twice, I think.

Writing this, I wonder if it would not be better to fix this by not passing `datetime` objects at all, just always stick to dates, but I guess you could also want to limit a range based on time too, so that would probably still need to work.

I previously said:

>  but it would be easy to support date as well (just do isinstance(start, date) instead of isinstance(start, pdt.date). Or would it be better not to support this? Or would the Range constructor already accept date objects properly maybe?

But that made no sense: The current `isinstance(start, pdt.date)` *will* already catch `date` instances, so it supports both `date` and `pdt.date` already.

ederag:
Range granularity is `datetime` only.

Otherwise, there would be ambiguities:
Suppose `range = Range(hday1, hday2)`, what is the type of `range.start` ?

Passing `hday` to `from_start_end()` is only a convenience, to ease transition.
I'd recommend an explicit `Range(hday1.start, hday2.end)` for new code.

GeraldJansen:
Free advice for new maintainers: This PR deals with a **serious** bug because it leads to incorrect invoices for anyone like me who relies on `hamster export` for invoicing. The bug should be fixed quickly and v3.0.2 should be released ASAP. Further code improvement can come later.

ederag:
Indeed. And the code is fine now.
It is ready to ship, or are there any concerns ?

mwilck:
Let's merge it, then. The 3.0.2 version bump should be discussed elsewhere.

PR #584, opened 2020-03-10
Use itstool instead of xml2po

### LINKED ISSUE #583: Don't use deprecated xml2po (gnome-doc-utils)
Hamster currently uses xml2po to generate localized help files. xml2po is part of the [gnome-doc-utils project](https://gitlab.gnome.org/Archive/gnome-doc-utils), which is deprecated and replaced by [itstool](http://itstool.org/) / [yelp-tools](https://gitlab.gnome.org/GNOME/yelp-tools).

Please replace the xml2po usage with itstool.

City-busz:
xml2po is deprecated and itstool from Yelp is recommended as its
replacement.

The code in wscript that handles this looks at the Makefile.am file to
figure out what pages and languages to compile. The Makefile is not
actually executed in this process, but is presumably present to allow
manually compiling the helpfiles too. This makefile is adapted to the
style used by itstool, changing DOC_ variables to HELP_ variables.

This also moves legal.xml from DOC_INCLUDES to HELP_PAGES, to make sure
it is still processed. At the same time, the python code is modified to
also process .xml files (instead of just .page files). Note that before
this commit, legal.xml would be processed by the Makefile, but not by
the wscript file, which is now fixed.

Whereas xml2po (when run with -p) would combine an XML file and a .po
file and produce a translated xml file, itstool does the same but with a
.mo file. This means that the .po file must be compiled into a .mo
beforehand, which done with an added msgfmt rule.

This fixes #583.

### DIFF
--- a/help/Makefile.am
+++ b/help/Makefile.am
@@ -1,11 +1,8 @@
-include $(top_srcdir)/gnome-doc-utils.make
-dist-hook: doc-dist-hook
+@YELP_HELP_RULES@
 
-DOC_ID = hamster
-DOC_INCLUDES = legal.xml
-DOC_LINGUAS = cs da de el es fa fr gl hu pl pt_BR ro ru sl te zh_CN zh_HK zh_TW
+HELP_ID = hamster
 
-DOC_PAGES = \
+HELP_FILES = \
         backup.page \
         index.page \
         input.page \
@@ -13,4 +10,7 @@ DOC_PAGES = \
         merge.page \
         reports.page \
         statistics.page \
-        tracking.page
+        tracking.page \
+        legal.xml
+
+HELP_LINGUAS = cs da de el es fa fr gl hu pl pt_BR ro ru sl te zh_CN zh_HK zh_TW
--- a/help/Makefile.am
+++ b/help/Makefile.am
@@ -1,6 +1,7 @@
 @YELP_HELP_RULES@
 
 HELP_ID = hamster
+HELP_LINGUAS = cs da de el es fa fr gl hu pl pt_BR ro ru sl te zh_CN zh_HK zh_TW
 
 HELP_FILES = \
         backup.page \
@@ -12,5 +13,3 @@ HELP_FILES = \
         statistics.page \
         tracking.page \
         legal.xml
-
-HELP_LINGUAS = cs da de el es fa fr gl hu pl pt_BR ro ru sl te zh_CN zh_HK zh_TW
--- a/help/wscript
+++ b/help/wscript
@@ -34,7 +34,7 @@ def apply_mallard(self):
 	lst = self.to_list(bld.env["HELP_LINGUAS"])
 	cnode = self.path.find_dir("C")
 
-	pages = [p.name for p in cnode.ant_glob("*.page")]
+	pages = [p.name for p in cnode.ant_glob(["*.page", "*.xml"])]
 
 	# Check if the declared page list is consistent
 	declared_pages = self.to_list(bld.env["HELP_FILES"])
@@ -54,10 +54,10 @@ def apply_mallard(self):
 		    target=mo,
 		    )
 		for page in pages:
-			src = self.path.find_resource('C/%s' % page)
 			out = self.path.find_or_declare('%s/%s' % (lang, page))
+			src = self.path.find_resource('C/%s' % page)
 			bld(name='itstool', color='BLUE',
-			    rule='${ITSTOOL} ${ITSTOOLFLAGS} ${SRC} -o ${TGT}',
+			    rule='${ITSTOOL} -m ${SRC} -o ${TGT}',
 			    source=[mo, src],
 			    target=out,
 			    install_path=install_path(lang)
@@ -85,7 +85,6 @@ def configure(ctx):
 		ctx.find_program('msgfmt', var='MSGFMT')
 		ctx.find_program(lookfor, var='ITSTOOL')
 		ctx.env.docs = True
-		ctx.env.ITSTOOLFLAGS = '-m'
 	except ctx.errors.ConfigurationError:
 		Logs.warn("'{}' not found; documentation build disabled".format(lookfor))
 		ctx.env.docs = False
--- a/help/wscript
+++ b/help/wscript
@@ -9,50 +9,56 @@ import os
 from waflib import Logs, TaskGen
 
 def _read_makefile_am(filename):
-	"read a Makefile.am file for DOC_* variable definitions, return a dict"
+	"read a Makefile.am file for HELP_* variable definitions, return a dict"
 	varstring = open(filename).read()
 	varstring = varstring.replace("\\\n", " ")
-	varlines = [L for L in varstring.splitlines() if L.startswith("DOC")]
+	varlines = [L for L in varstring.splitlines() if L.startswith("HELP")]
 	return dict(tuple(map(str.strip, var.split("=", 1))) for var in varlines)
 
 def init_mallard(self):
 	mf_am = self.path.find_resource(self.variable_definitions)
-	DOC_VAR = _read_makefile_am(mf_am.abspath())
+	HELP_VAR = _read_makefile_am(mf_am.abspath())
 
-	require_vars = "DOC_ID DOC_LINGUAS DOC_PAGES".split()
-	have_vars = set(var for var in DOC_VAR if DOC_VAR[var])
+	require_vars = "HELP_ID HELP_LINGUAS HELP_FILES".split()
+	have_vars = set(var for var in HELP_VAR if HELP_VAR[var])
 	missing_vars = set(require_vars).difference(have_vars)
 	if missing_vars:
-		print("Missing DOC variable declarations in {}:".format(mf_am.abspath()))
+		print("Missing HELP variable declarations in {}:".format(mf_am.abspath()))
 		print("\n".join(missing_vars))
 
-	self.bld.env.update(DOC_VAR)
+	self.bld.env.update(HELP_VAR)
 	self.default_install_path='${PREFIX}/share/help'
 
 def apply_mallard(self):
 	bld = self.bld
-	lst = self.to_list(bld.env["DOC_LINGUAS"])
+	lst = self.to_list(bld.env["HELP_LINGUAS"])
 	cnode = self.path.find_dir("C")
 
 	pages = [p.name for p in cnode.ant_glob("*.page")]
 
 	# Check if the declared page list is consistent
-	declared_pages = self.to_list(bld.env["DOC_PAGES"])
+	declared_pages = self.to_list(bld.env["HELP_FILES"])
 	missing_pages = set(pages).difference(declared_pages)
 	if missing_pages:
 		print("Warning: Some pages not declared:")
 		print("\n".join(missing_pages))
 
-	install_path = lambda lang: os.path.join(bld.env.DATADIR, "help", lang, "${DOC_ID}")
+	install_path = lambda lang: os.path.join(bld.env.DATADIR, "help", lang, "${HELP_ID}")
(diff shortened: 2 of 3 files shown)


help/wscript
```diff
@@ -74,11 +80,12 @@ TaskGen.after('init_mallard')(apply_mallard)
 
 
 def configure(ctx):
-	lookfor = "xml2po"
+	ctx.find_program('msgfmt', var='MSGFMT')
```
benjaoming: The below check for `itstool` crashes and gently disables documentation, but the check for `msgfmt` will stop the build. Is that intended? I'm not even sure why it should be normal to build without the translations, since it's clearly stated as a build dependency in the README -- not sure if other assumptions break later on, anyways?

Seems that help, documentation and translation is a bit mixed up, too.

Could the whole exception handling below be removed or does it have a reason to be?

City-busz: I don't have too much experience with waf. Now I moved it inside the exception handling.

mwilck:
So according to our new rules, with two approvals, we'd be able to merge, right?

benjaoming:
@mwilck - yes, I can see a green Merge button, can you?

help/wscript
benjaoming: Thanks for addressing it :+1:

help/Makefile.am
```diff
@@ -1,16 +1,16 @@
...
         backup.page \
         index.page \
         input.page \
         keys.page \
         merge.page \
         reports.page \
         statistics.page \
-        tracking.page
+        tracking.page \
+        legal.xml
```
matthijskooijman: This seems to add an additional `legal.xml`, which seems unrelated to the xml2po -> itstool change? If so, it might be better in a separate commit.

help/Makefile.am
```diff
@@ -1,16 +1,16 @@
...
+
+HELP_LINGUAS = cs da de el es fa fr gl hu pl pt_BR ro ru sl te zh_CN zh_HK zh_TW
```
matthijskooijman: Why move this line down?

help/Makefile.am
matthijskooijman: Ah, I guess it was previously automatically added somehow through the `DOC_INCLUDES` variable?

help/Makefile.am
```diff
@@ -1,16 +1,16 @@
-include $(top_srcdir)/gnome-doc-utils.make
-dist-hook: doc-dist-hook
+@YELP_HELP_RULES@
```
matthijskooijman: What does this line do exactly? AFAICS we're not actually executing the Makefile, only parsing it to get some variables? Is this needed then? Or does it allow manually building with make?

help/wscript
```diff
@@ -9,50 +9,56 @@ import os
...
+	install_path = lambda lang: os.path.join(bld.env.DATADIR, "help", lang, "${HELP_ID}")
 
 	for lang in lst:
 		node = self.path.find_resource("%s/%s.po" % (lang, lang))
+		mo = self.path.find_or_declare("%s/%s.mo" % (lang, lang))
+		bld(name='msgfmt', color='BLUE',
+		    rule='${MSGFMT} ${SRC} -o ${TGT}',
+		    source=node,
+		    target=mo,
+		    )
```
matthijskooijman: Why is this msgfmt call added? It seems weird to me that this was not previously needed (as xml2po probably produces .po files, so I would expect a msg2fmt (which I think converts .po to .mo) call with xml2po, not with `itstool` (which I suspect converts to .mo directly?)

help/Makefile.am
```diff
@@ -1,16 +1,16 @@
...
 
-DOC_ID = hamster
-DOC_INCLUDES = legal.xml
-DOC_LINGUAS = cs da de el es fa fr gl hu pl pt_BR ro ru sl te zh_CN zh_HK zh_TW
+HELP_ID = hamster
 
-DOC_PAGES = \
+HELP_FILES = \
```
matthijskooijman: Why rename all this stuff from `DOC_` to `HELP_`, is this because the Yelp makefile rules needs that?

help/Makefile.am
City-busz: It's not really needed. It's just needed if someone wants to use autotools.

help/Makefile.am
City-busz: Yes, these are the variables used by [yelp-tools](https://wiki.gnome.org/Apps/Yelp/Tools/yelp.m4). It's not really needed until autotools are not used, it's just for compatibility.

help/Makefile.am
City-busz: Yes, it's just a compatibility with yelp-tools, related to other variable renames.

help/Makefile.am
City-busz: I just replaced the variables from yelp-tools, it's just moved accidentally.

help/wscript
City-busz: It seems to me that itstool expects .mo files, and can't handle .po files directly. [The reference](http://itstool.org/documentation/basic-usage.html#merge) says that the po files need to be converted to mo files first.

help/wscript
```diff
@@ -9,50 +9,56 @@ import os
...
 		for page in pages:
-			out = self.path.find_or_declare('%s/%s' % (lang, page))
 			src = self.path.find_resource('C/%s' % page)
-			bld(name='xml2po', color='BLUE',
-			    rule='${XML2PO} ${XML2POFLAGS} ${SRC} > ${TGT}',
-			    source=[node, src],
+			out = self.path.find_or_declare('%s/%s' % (lang, page))
```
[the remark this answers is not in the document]
City-busz: Yes, it could stay at the top.

mwilck:
@matthijskooijman, do the replies from @City-busz satisfy you? The move of the `out=` line wouldn't be a reason to reject this IMO.

benjaoming:
@mwilck yes, it also runs fine :+1:

help/wscript
matthijskooijman: Ah, I see how it works now. Indeed, xml2po was previously used not to generate po files, but it *read* the po file and produced a translated XML. The bld commandline here is a bit confusing, because the -p option (for read .po file and translate xml) was hidden in XML2POFLAGS, and source contains *two* filenames, which are both passed in the place of `${SRC}`. So the old code would take XML and .po and produce translated xml. The new itstool takes XML and .mo and produces translated XML. So, this is clarified.

help/Makefile.am
matthijskooijman: I wonder if that is still relevant, but removing this is probably out of scope for this PR, so good to keep it working for now.

help/Makefile.am
matthijskooijman: Ok, thanks for clarifying.

help/Makefile.am
matthijskooijman: Hm, looking at the old [`xml2po` Makefile include](https://gitlab.gnome.org/GNOME/ekiga/blob/master/gnome-doc-utils.make), it indeed seems that `DOC_INCLUDES` was pretty much included in the translation just like the help pages, so this change makes sense. However, it seems that the waf-based processing of translations did *not* actually look at `DOC_INCLUDES`, so quite possibly this file was not processed with `xml2po` before and this fix fixed by this commit.
*however*, I tried running this just now, and it seems it's still not processed by itstool somehow. I haven't figured out why yet...

help/wscript
```diff
@@ -9,50 +9,56 @@ import os
...
+	self.bld.env.update(HELP_VAR)
 	self.default_install_path='${PREFIX}/share/help'
 
 def apply_mallard(self):
 	bld = self.bld
-	lst = self.to_list(bld.env["DOC_LINGUAS"])
+	lst = self.to_list(bld.env["HELP_LINGUAS"])
 	cnode = self.path.find_dir("C")
 
 	pages = [p.name for p in cnode.ant_glob("*.page")]
```
matthijskooijman: This line prevents legal.xml from being processed.

help/Makefile.am
matthijskooijman: Ah found it, it only processes *.pages. Added a comment in the relevant place and resolving this comment.

matthijskooijman:
I found one more issue (legal.xml not processed), see inline comment, that I think should be fixed before merging. Also, I'd like to see the commit messages expanded with some more info, but I'm happy to do that tomorrow.

mwilck:
> I found one more issue (legal.xml not processed), see inline comment, that I think should be fixed before merging. 

How do you want to fix it - rename the file from `legal.xml` to `legal.page`?
IIUC @City-busz doesn't need to make additional changes, right?

benjaoming:
@mwilck Sounds great if @matthijskooijman does that other fix separately :+1:

mwilck:
@benjaoming: @matthijskooijman can apply minor changes later, but if he intends to edit the commit descriptions, we can't merge now. Or am I missing something?

matthijskooijman:
> How do you want to fix it - rename the file from legal.xml to legal.page?

Easiest would be to just include *.xml in the glob in the python code, so I'll implement that in a minute.

Looking at this, I wonder if it wouldn't be better to just clean up this code even more and just always compile al present *.page *.xml files, without having a list (and if we want to keep the makefile supported, that can also just do the same wildcard), but let's leave that out of scope of this PR.

mwilck:
> let's leave that out of scope of this PR.

Ack.

matthijskooijman:
I've pushed some more changes. I've used an additional fixup commit (rather than amending the current commit), for easier review. When merging, we should select "Squash and merge" to make sure this fixup commit does not end up in the actual history.

In the commit, I moved back the lines that did not need moving, fixed the legal.xml thing and also slightly clarified the itstool command (rather than hiding the -m option in an ITSTOOLFLAGS, it is not specified explicitly so it is more clear that the first filename from `SRC` is passed to the -m flag. I tried moving the .mo filename into a separate `MOFILE` variable, but could not get that to work quickly, so I left it at this.

I also proposed a commit message in the first post of this PR. I was hoping that github would automatically use this message in the merge commit description, but it does not seem to, so when merging we should copy the message from the first post into the "merge" input box I think (or squash the commits manually first and then do a regular merge maybe).

Does this look ok like this?

mwilck:
> I also proposed a commit message in the first post of this PR

Not sure I understand - do you mean [this post](https://github.com/projecthamster/hamster/pull/584#issue-386155626)?

benjaoming:
Nice, glad we got @City-busz 's contribution in, well done @City-busz :tada:

> for easier review

Yes, that's very nifty @matthijskooijman to amend things on top of already reviewed code - also nice to keep @City-busz 's commits for full attribution / origin etc.

benjaoming:
@matthijskooijman 

I have copy-pasted the description of the PR into the merge commit -- hoping that adds the git history that you meant?

mwilck:
Yes, thanks a lot @City-busz :+1:

mwilck:
@benjaoming, I thought you wanted to use "merge and squash" to get rid of the "fixup!" commit (as you said you prefer clean commit history)... but that's just a nit. :rocket:

matthijskooijman:
Yay for merging this. Merge commit message is as I intended, perfect!

However, it seems you did not squash and merge as I intended, so now we have this "fixup" commit in the history (that really should have been squashed into the first commit - no real point in having this as two commits in the history, especially since part of the second commit was intended to reduce noise (and also blame pollution) on the first commit). There is a little point in keeping the second commit separate to preserve authorship, but then the commit message should have been nicer (and in this case, the changes are so small that it did not seem relevant).

benjaoming:
Okay - I get what you mean now: to have the fixup commit separate in the PR but squashed in the `master` branch :+1:

matthijskooijman:
Exactly! :-)

Issue #585, opened 2020-03-11
Project policies


matthijskooijman:
With our new maintainer team, I think it would be good to think a bit on how to approach this, and what policies we should adapt. This was already previously discussed a bit, so here's a proposal for working with issues and PRs:
 - All changes should go through PRs. Direct (and certainly force) pushes to master are not allowed (and also technically prevented by branch protection).
 - A PR can be merged when it has two accepting reviews. This includes the person doing the merge, but can also include reviews from community members without repo write access. For small or trivial changes, just one review can be sufficient. Also, for a PR from the maintainer team, just one review can also be sufficient (to allow a PR to be merged with just two maintainers involved).
 - PRs should have clear commit messages and commits should be clean and atomic (not mixing up different changes in the same commit). Rebasing and history editing of PRs is encouraged to keep the history clean (but sometimes *adding* fixup commits first and then squashing commits just before merge can be easier for review).
 - Issues should mostly be about bugs and feature requests and should remain open as long as the bug is present or there is intention to (someday) implement a feature. Issues that ask for help with a problem or ask questions are ok, but should usually be closed rather quickly (discussion can still continue on a closed issue anyway).
 - Issues that are previously reported, should be closed with a comment saying they're duplicate and pointing to the previous comment.

Something like that?

mwilck:
ACK for all items. Thanks for writing this up.

Wrt the commit messages, you are introducing a culture change here. ederag's commit messages used to be terse to the extreme. But I agree it's a change for the better. We just have to make sure we don't exaggerate.

ederag:
> ederag's commit messages used to be terse to the extreme

Indeed, information was given as comments/docstrings in the code itself.

mwilck:
good to know that you're still listening :-)

matthijskooijman:
I added some more details on handling PRs and added some links for extra documentation.

> Wrt the commit messages, you are introducing a culture change here. ederag's commit messages used to be terse to the extreme. But I agree it's a change for the better. We just have to make sure we don't exaggerate.

It's a change, but I'm not sure it is too big. I think @ederag has rather concise style of commit messages, and my personal style is quite verbose, but I would not want to impose my verbosity on the project (above, I only wrote "clear" wrt commit messages).

Wrt to clean and atomic commits, this is already the culture of this repo - @ederag has done a great job of cleanly separating unrelated changes and refactors from actual fixes.

ederag:
[...]

PR #596, opened 2020-04-11
Daily summary totals added to Overview screen

esteban-aliverti:
While migrating to Ubuntu 20.04 from 16.04 I noticed that the current version of Hamster doesn't provide daily totals in the Overview screen. This PR introduces a "Total" row after each day. These rows can't be selected (clicked) and navigation with the cursor keys skip them. These rows are not included when exporting the tasks.

Here you can find a screenshot of how the modifications look like: https://ibb.co/GFPv2Sj

Please note that I'm not really familiar with Python, so please let me know if there is something in my changes that can be improved.

### DIFF
--- a/src/hamster/widgets/facttree.py
+++ b/src/hamster/widgets/facttree.py
@@ -30,6 +30,7 @@ from gi.repository import Pango as pango
 from hamster.lib import datetime as dt
 from hamster.lib import graphics
 from hamster.lib import stuff
+from hamster.lib.fact import Fact
 
 
 class ActionRow(graphics.Sprite):
@@ -45,6 +46,20 @@ class ActionRow(graphics.Sprite):
 
         self.width = 50 # Simon says
 
+class TotalFact(Fact):
+   """An extension of Fact that is used for daily totals.
+   Instances of this class are rendered differently than instances
+   of Fact.
+   A TotalFact doesn't have a meaningful  start and an end, but a 
+   total duration (delta).
+   """
+   def __init__(self, activity, duration):
+        super().__init__(activity=activity, start=dt.datetime.now(), end=dt.datetime.now())
+        self.duration = duration
+
+   @property
+   def delta(self):
+        return self.duration
 
 
 class Label(object):
@@ -66,6 +81,9 @@ class Label(object):
     def set_text(self, text):
         self.text = text
         self.layout.set_markup(text)
+
+    def get_text(self):
+        return self.text
 
     def show(self, g, text=None, x=None, y=None):
         """Show the label.
@@ -206,8 +224,11 @@ class FactRow(object):
         g.translate(self.row_margin_H, self.row_margin_V)
 
         g.set_color(color)
-        self.time_label.show(g)
-        self.activity_label.show(g)
+
+        #Do not show the start/end time for Totals
+        if not isinstance(self.fact, TotalFact):
+            self.time_label.show(g)
+        self.activity_label.show(g, self.activity_label.get_text() if not isinstance(self.fact, TotalFact) else "<b>{}</b>".format(self.activity_label.get_text()))
 
         if self.fact.category:
             g.save_context()
@@ -233,7 +254,7 @@ class FactRow(object):
 
             g.restore_context()
 
-        self.duration_label.show(g, self.fact.delta.format(), x=self.width - 105)
+        self.duration_label.show(g, self.fact.delta.format() if not isinstance(self.fact, TotalFact) else "<b>{}</b>".format(self.fact.delta.format()), x=self.width - 105)
 
         g.restore_context()
 
@@ -252,9 +273,10 @@ class FactTree(graphics.Scene, gtk.Scrollable):
 
 
     ASCII Art!
-    | Weekday    | Start - End | Activity - category   [actions]| Duration |
-    | Month, Day |             | tags, description              |          |
-    |            | Start - End | Activity - category            | Duration |
+    | Weekday    | Start - End | Activity - category   [actions]| Duration       |
+    | Month, Day |             | tags, description              |                |
+    |            | Start - End | Activity - category            | Duration       |
+    |            |             | Total                          | Total Duration |
 
     Inline edit?
 
@@ -331,7 +353,8 @@ class FactTree(graphics.Scene, gtk.Scrollable):
                     and self.hover_fact.id == self.current_fact.id)
                ):
                 self.unset_current_fact()
-            else:
+            #Totals can't be selected
+            elif not isinstance(self.hover_fact,TotalFact):
                 self.set_current_fact(self.hover_fact)
 
     def activate_row(self, day, fact):
@@ -341,7 +364,7 @@ class FactTree(graphics.Scene, gtk.Scrollable):
         self.emit("on-delete-called", fact)
 
     def on_double_click(self, scene, event):
-        if self.hover_fact:
+        if self.hover_fact and not isinstance(self.hover_fact,TotalFact):
             self.activate_row(self.hover_day, self.hover_fact)
 
     def on_key_press(self, scene, event):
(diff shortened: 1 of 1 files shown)


GeraldJansen:
Nice. Note that your totals are incorrect (eg. on Tuesday in your example) because they don't account for untracked breaks. You probably should loop over the facts for each day and add up the individual durations. Furthermore, it would be be good to avoid showing the total when displaying a single date (len(by_date) == 1?) because it would be redundant.

esteban-aliverti:
Good point about the incorrect totals. I'll need to change a little bit the code, because I can't use a Fact anymore to keep the total, but It should be easy to do. 
Regarding hiding the total for days with one activity, I would like to hear what others have to say. I'm showing the total on purpose to have symmetry. I guess that I can hide it when displaying a single day, but still show them for weekly or monthly filters. Thoughts?

esteban-aliverti:
Here are 2 screenshots with the incorrect totals bug fixed. 

This screenshot here - https://ibb.co/WxNNzc7 - has the "Total" row always enabled.
This other screenshot here - https://ibb.co/zFV9YNy - only shows the "Total" for days with more than 1 activity.

I really like the first approach, because it is coherent: you know there is a bold row with the totals after each day. But I understand that this is a personal preference, and I don't have any problem in leave it that way if that is what you prefer. 

As soon as I get a response from you I'll be updating this PR.

Thanks!

matthijskooijman:
Personally, I think I would prefer showing the row only for days with multiple activities, though always showing might also be easier to read, so it's a close call for me.

How are you handling single-day views now? I guess those do not need the total row, since there is already a (day) total in bottom bar then?

GeraldJansen:
Regarding showing the totals, I agree with "hide it when displaying a single day, but still show them for weekly or monthly filters" [or any custom multi-day period]. That's what I was suggesting, perhaps not clearly. When showing multiple days I agree that showing the total even for days with a single task is more consistent.

esteban-aliverti:
That makes sense. I'll implement it that way and update this PR. Thank you for your prompt response.

esteban-aliverti:
So, this is how it looks now:

*  Single day: https://ibb.co/0MpdQ3P (No "Total" row) 
*  Multi-day: https://ibb.co/kHqLss6 ("Total" row is visible)

Please let me know your opinion.

matthijskooijman:
Looks good to me. I have not looked at the code (yet).

src/hamster/widgets/facttree.py
```diff
@@ -45,6 +46,20 @@ def __init__(self):
 
         self.width = 50 # Simon says
 
+class TotalFact(Fact):
```
mwilck, line 49: I find it awkward that this class inherits from "Fact" given that it "_doesn't have a meaningful  start and an end_".

esteban-aliverti: Yeah, I made it a subclass of Fact so I didn't have to change some of the methods in the `FactTree` class. Some of the methods in this class were already dealing with arrays or instanced of the `Fact` class. Making the Total row a Fact made it easier (in my opinion) to introduce the changes pursued by this PR. If you prefer, I can make the `TotalFact` not to extend from `Fact`, but the code in `FactTree` will become messier (again, in my opinion).

mwilck: Sure, it was easier this way :grin:. I wondered if it'd be possible to separate out the functionality that "TotalFact" needs from "Fact" into another base class, and then have them both inherit from that new base class. But I haven't looked deep enough to be certain that that would be feasible.

I guess I'd be fine with just a comment with a FIXME saying that one day, the class hierachy needs to be cleaned up.

esteban-aliverti: I'll definitely will add a comment. I was also thinking about creating a superclass for both classes, but I didn't want to mess around too much with the code in my first PR! :)

esteban-aliverti:
I added the **FIXME** message. Do you guys prefer squashed commits for PRs or no?

ederag:
Former maintainer here.
The plan was to rework facttree. That's the meaning of the [note](https://github.com/projecthamster/hamster/projects/1#card-25578550):
> Fully exploit that FactTree is a painter, never modifies facts by itself, to clarify code.

It would be much clearer to loop on facts,
paint facts along the way, 
and paint the total and the date upon day change.
No need to hack on `Fact` or add a new class.

esteban-aliverti:
@ederag thanks for your input. But what you are suggesting would require a major (or at least big) rework, right? As far as I understand, there are 3 operations in `FactTree` involved in the rendering of a Fact:
1.- The list of Facts is set using the `set_facts()` method. The `FactTree` class keeps a copy of these facts for later use.
2.- Different methods in the `FactTree` class calculate the facts that are visible at any time (`visible_range`).
3.- The `on_enter_frame()` method draws each of the **visible** FactRows by calling the `FactRow.show()` method. 

In this flow, there only place where we have all the facts for a single day (and the total can be calculated) is in the `set_facts()` method. When the facts are being painted it seems to be too late, because the code is only dealing with the subset of visible facts.
Also, given that a `FactRow` expects an instance of `Fact` (`FactRow.set_fact()`), that would mean that we will still need a subclass of `Fact`, or a subclass of `FactRow` in any case. 

Am I missing something? 

Thanks again for your insights.

ederag:
> what you are suggesting would require a major (or at least big) rework

Exact. Your proposal makes perfect sense for a new contributor
(minimal changes to achieve a precise goal, lighter to review).
But for the long run maintainability, a rework could also make sense.

Remember I'm a normal contributor now. Below are just ideas.

Your assessments seems right.
Indeed, totals have to be evaluated in `set_facts()`.
A special class, say `TotalRow` could paint the totals.
`TotalRow` and `FactRow` might inherit from a master `RowPainter` class (if that removes duplication).
Note: just as Facttree is using a single FactRow painter instance,
a single `TotalRow` instance would paint the totals.
(side note: `FactTreePainter `, `FactRowPainter` and `TotalRowPainter` 
 might be better names, more explicit)

Since the whole thing is a painter, and facts are modified only by `set_facts`,
the number of rows is set there. So it would be easier and more readable to have a few lists,
like the existing `row_positions` and `row_heights`: 
`row_contents` (elements are `Fact` or `timedelta`)
maybe change  `days` to `row_days` (day the row belongs to).

Meddling with facts should be forbidden, especially adding attributes (`.y`).

`get_visible_range()` would just yield indices of visible rows,
much simpler than its current output.

`on_enter_frame()` would loop on these indices,
and call either `self.fact_row.set_fact()` or `self.total_row.set_duration()` 
depending on the `row_contents` element type.

Please take all this with a grain of salt; it's been a while since I looked at this code in details.

esteban-aliverti:
Hi there, 
At this point I'm not sure if you are waiting for me to change something in the PR, if you want to close it because you want to wait for the refactors that @ederag mentioned or if it's just that you are busy and didn't have time to review the changes. 

Thanks in advance,

esteban-aliverti:
Any news on this PR?

benjaoming:
Just read through the PR and it seems that all the feedback has been addressed. Of course would be nice with the class refactor, but if that was to happen, this would also be hard to release in a patch release 3.0.2.

Just tested and it works smoothly, thanks @esteban-aliverti for addressing all the feedback :100: 

Seeing the scope of the changes, I think it looks good for 3.0.2, so have added that milestone and tagged this PR.

esteban-aliverti:
Fantastic! Keep on with the great job you are doing guys!

matthijskooijman:
> Seeing the scope of the changes, I think it looks good for 3.0.2, so have added that milestone and tagged this PR.

I'm going to disagree here: I'd rather go ahead and release 3.0.2 soon without this (I wrote "now" here before, but ran out of time for today again...).

Since this is a new feature, I'd rather merge this and have it sit in master for a little bit to get some additional test exposure before including it into a release, rather than merging it and releasing it right away.

Also, is this PR completely ready to merge? You mentioned all feedback has been addressed, but have not given an approving review (and we need two before merging).

benjaoming:
> we need two before merging

We do? I don't see any warnings.

src/hamster/widgets/facttree.py
benjaoming: I really understand that you don't want to mess around with a lot of code in the first PR, @esteban-aliverti :)

benjaoming:
Merging this now that 3.0.2 has been released, great work releasing that @matthijskooijman :tada:

matthijskooijman:
> We do? I don't see any warnings.

That's what we discussed at #585 (but it's not enforced, since it's a general guideline, not a hard rule). Regardless, no objections from my side, everyone has had more than enough time to review if they really wanted (including me) :-)

benjaoming:
Aaah right right, thanks for reminding. The rule is good. I guess the overall feeling of whether something is duly and properly reviewed will always present itself in some shape or form :)

PR #597, opened 2020-04-17
Fix date

### LINKED ISSUE #590: Offset between calendar selection and cmdline
I found that Hamster is adding some offset between dates selected via the calendar and those that land in "cmdline". For example, if I edit a task started and finished on 15th and change the date to 18th with the calendar, the command will start with "2020-03-21". This is very confusing.

image

ederag:
Should fix #590.
This one has been unstaged for about three weeks
(my motherboard crashed, too old, had to change the whole computer, notably).

No warranty, that's untested.

### DIFF
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -248,7 +248,9 @@ class CustomFactController(Controller):
                     # preserve fact duration
                     self.fact.end_time += delta
                     self.end_date.date = self.fact.end_time
-            self.day = self.fact.date or dt.hday.today()
+            if self.fact.date:
+                # change default day only if date is set
+                self.day = self.fact.date
             self.validate_fields()
             self.update_cmdline()
 
@@ -268,7 +270,7 @@ class CustomFactController(Controller):
                                                          new_time)
                 else:
                     # date not specified; result must fall in current hamster_day
-                    new_start_time = dt.datetime.from_day_time(dt.hday.today(), new_time)
+                    new_start_time = dt.datetime.from_day_time(self.day, new_time)
             else:
                 new_start_time = None
             self.fact.start_time = new_start_time
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -146,10 +146,10 @@ class CustomFactController(Controller):
             # self.update_fields() here would enter an infinite loop
 
     def on_prev_day_clicked(self, button):
-        self.increment_date(-1)
+        self.increment_day(-1)
 
     def on_next_day_clicked(self, button):
-        self.increment_date(+1)
+        self.increment_day(+1)
 
     def draw_preview(self, start_time, end_time=None):
         day_facts = runtime.storage.get_facts(self.day)
@@ -159,7 +159,7 @@ class CustomFactController(Controller):
         """ skip one variable (huh) """
         return self._gui.get_object(name)
 
-    def increment_date(self, days):
+    def increment_day(self, days):
         delta = dt.timedelta(days=days)
         self.day += delta
         self.update_fields()
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -101,7 +101,7 @@ class CustomFactController(Controller):
 
         original_fact = self.fact
         # TODO: should use hday, not date.
-        self.date = self.fact.date
+        self.day = self.fact.date
 
         self.update_fields()
         self.update_cmdline(select=True)
@@ -129,12 +129,12 @@ class CustomFactController(Controller):
         self.window.show_all()
 
     @property
-    def date(self):
+    def day(self):
         """Default hamster day."""
         return self._day
 
-    @date.setter
-    def date(self, value):
+    @day.setter
+    def day(self, value):
         delta = value - self._day if self._day else None
         self._day = value
         self.cmdline.default_day = value
@@ -152,8 +152,8 @@ class CustomFactController(Controller):
         self.increment_date(+1)
 
     def draw_preview(self, start_time, end_time=None):
-        day_facts = runtime.storage.get_facts(self.date)
-        self.dayline.plot(self.date, day_facts, start_time, end_time)
+        day_facts = runtime.storage.get_facts(self.day)
+        self.dayline.plot(self.day, day_facts, start_time, end_time)
 
     def get_widget(self, name):
         """ skip one variable (huh) """
@@ -161,7 +161,7 @@ class CustomFactController(Controller):
 
     def increment_date(self, days):
         delta = dt.timedelta(days=days)
-        self.date += delta
+        self.day += delta
         self.update_fields()
 
     def show(self):
@@ -186,7 +186,7 @@ class CustomFactController(Controller):
 
     def on_cmdline_changed(self, widget):
         if self.master_is_cmdline:
-            fact = Fact.parse(self.cmdline.get_text(), default_day=self.date)
+            fact = Fact.parse(self.cmdline.get_text(), default_day=self.day)
             previous_cmdline_fact = self.cmdline_fact
             # copy the entered fact before any modification
             self.cmdline_fact = fact.copy()
@@ -249,7 +249,7 @@ class CustomFactController(Controller):
                     # preserve fact duration
                     self.fact.end_time += delta
                     self.end_date.date = self.fact.end_time
-            self.date = self.fact.date or dt.hday.today()
+            self.day = self.fact.date or dt.hday.today()
             self.validate_fields()
             self.update_cmdline()
 
@@ -289,7 +289,7 @@ class CustomFactController(Controller):
     def update_cmdline(self, select=None):
         """Update the cmdline entry content."""
         self.cmdline_fact = self.fact.copy(description=None)
-        label = self.cmdline_fact.serialized(default_day=self.date)
+        label = self.cmdline_fact.serialized(default_day=self.day)
         with self.cmdline.handler_block(self.cmdline.checker):
             self.cmdline.set_text(label)
             if select:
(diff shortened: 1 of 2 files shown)


mwilck:
Can we get this in small, easy-to-review pieces, please? For example the fix for #590 separated out?

ederag:
@mwilck Rebased on master.
Disclaimer: only a simple check has been done, please test thoroughly.

mwilck:
I'm missing an explanation why this this fixes the problem. The renaming of the methods / fields doesn't help to fix anything, AFAICS. The last commit is the only one that might consitiute a fix, although I don't see why it would. My analysis after playing with this code is https://github.com/projecthamster/hamster/issues/590#issuecomment-618707131, and I don't understand why this PR would change anything about the double-application of the time delta I saw there.

ederag:
Indeed, the first commits just clarified code, enough to find the simple bug fix (or so it seems ?).
Please test first, and report back.

src/hamster/edit_activity.py
```diff
@@ -329,30 +330,31 @@ def validate_fields(self):
...
-        Try to merge date, activity and description informations.
+        Try to merge day, activity and description informations.
 
         Return the consolidated fact if successful, or None.
         """
         fact = self.fact
 
         now = dt.datetime.now()
-        self.get_widget("button-next-day").set_sensitive(self.date < now.date())
+        today = now.hday()
```
benjaoming, line 340:
```
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> datetime.now
<built-in method now of type object at 0xa346e0>
>>> datetime.now()
datetime.datetime(2020, 4, 27, 13, 16, 56, 954984)
>>> datetime.now().hday()
  File "<stdin>", line 1, in <module>
AttributeError: 'datetime.datetime' object has no attribute 'hday'
```

ederag: What this theoretical, or based on an actual test ?
Please look at the imports: `from hamster.lib import datetime as dt`

```bash
~/share/prog/python/hamster/src>GSETTINGS_SCHEMA_DIR=../build/data   python3 
Python 3.6.10 (default, Jan 16 2020, 09:12:04) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from hamster.lib import datetime as dt
>>> now = dt.datetime.now()
>>> now.hday()
hday(2020, 4, 27)
```

benjaoming: ah okay, that's where the object came from! I didn't recognize `hday`, was just trying to be helpful, thinking this was part of the untested code :)

matthijskooijman:
I just checked out this PR and did a quick test, the last commit does seem to fix the problem for *new* facts, but not when editing existing facts (which does seem to make sense given the code changed). So, this is not actually a complete fix, and probably not the right fix either?

ederag:
What are you doing exactly ?
When I edit an old fact from the april 21st, and change date (with the timeline arrows),
the start and end dates follow correctly here.

src/hamster/edit_activity.py
ederag: > Maybe some renaming could help here at some point.

And forget about compatibility with existing external code ?

matthijskooijman:
> When I edit an old fact from the april 21st, and change date (with the timeline arrows), the start and end dates follow correctly here.

The problem is not the timeline arrows, those work for me just fine in master already. The problem is when editing the start date using the calendar underneath  the "start" field. When I open up an existing item (I tried with one of today and yesterday), then open up the calendar and change the date, then the timeline (and I presume default date) get updated correctly, but the cmdline shows a start date which has the change applied twice.

This happens on existing *and* new items with master, and just with existing items with your PR applied.

matthijskooijman:
> This happens on existing and new items with master, and just with existing items with your PR applied.

Hm, I can no longer reproduce this: The bug now also occurs with new items with your PR applied. Not sure why that part seemed fixed before...

ederag:
2nd part *seems* fixed in https://github.com/ederag/hamster/tree/fix-calendar-offset.
Cherry-pick from it, *if you wish*.
This PR branch is yours now, so you may force-push to [ederag:fix-timeinput](https://github.com/ederag/hamster/tree/fix-timeinput).
but please do not try to create any PR on my side. [I'm out](https://github.com/projecthamster/hamster/issues/590#issuecomment-621068164).
It's *not* a fork; the `master` branch is lagging behind to keep that clear.

matthijskooijman:
#630 has a better fix for this issue, so I'm going ahead and close this, so we can focus on that fix instead.

matthijskooijman:
Let me further clarify why I closed this PR:
 - The PR as it is, does not fix the original problem completely, and as such is not ready to be merged.
 - The PR contains half a dozen commits and it is not immediately clear to me from the commit messages what they do exactly and why. There is no overall documentation about what the rationale behind this fix is.
 - @ederag has added a link to a branch with additional commits that supposedly do fix the problem, but again, it is unclear to me how the fix works exactly and I'm not so motivated to reverse engineer the fix (then I'd rather just spend that time to understand the code and come up with a fix myself).
 - There is another PR in #630 that does fix the original problem and has IMHO easier to understand and review commits.

I realize that #630 is not yet ready, since it introduces some new problems, as pointed out by @ederag, but I still believe that fixing #630 is a better approach than digging into the commits in this PR and the linked branch, so I still think this PR should be closed and there is no real point in referring back to this PR. If @ederag (or anyone else, for that matter) wants to clean up their fix and submit it as a proper, well-documented, easy-to-review PR, then that's of course fine, but that will still leave this PR closed.

ederag:
Perhaps not complete, but according to my tests, it does fix the main issue.

The correct stance would have been to either test it right away, 
or to take it as written by me, the former maintainer, and merged before your arrival.

All the first commits made are clarifications of my intent, as already said in https://github.com/projecthamster/hamster/pull/597#issuecomment-618937759
(that part code was work in progress, 
for instance some was written before we introduced the clear separation between civil date and hamster day), 
so there's no reverse engineering to do.
This makes this PR a much better start for any further improvement.

matthijskooijman:
I've been reconsidering the code in this PR and the branch linked by @ederag. To keep discussion in the right place, I'm moving some comments from #630 that relate to this.

I wrote:

> I also finally had a good look at @ederag's https://github.com/ederag/hamster/commits/fix-calendar-offset branch, and now I understand better what they are doing there. In fact, that branch already updates some uses of the hamsterday to use day (I was thinking about hday, but I guess day is already distinct enough from date, so that would be fine), so it seems we're in agreement about that part. I'm still having trying to puzzle out the reasons for some of those changes (lacking detailed commit messages), but I think the approach of renaming some things for clarity and then making some fixes make sense.

> So, I'm going to try to make a new PR for this issue, based on @ederag's branch, adding more detailed commit messages, restructuring some of the commits and maybe doing a few things different (especially in the later commits).

Then @ederag  wrote:

> Good that you eventually reconsidered #597. You got it.

> But I explicitly ask you not to tamper with my commits.
That's not OK. Their dates and styles are important. That's what future devs will see.
Cherry-pick or rebase as is,
then you may append your own commits, with your own style.

matthijskooijman:
> Their dates and styles are important. That's what future devs will see.

I agree that the commit separation, messages and message style is important, which is why I want to make some small changes to the separation and mostly make the commit messages more verbose, so those future devs will have an easier time understanding these changes. I'll also take care to preserve commit dates where appropriate.

If you insist that you do not want your commits to be modified in anyway (other than rebasing or cherry-picking), I do not consider them usable as-is and I would apparently need to create my own commits (some of which will end pretty much identical to yours). Where appropriate (i.e. for code copied from your commits) I'd credit you in the commit message, but not in the commit Author field then. If you prefer this, then I'll oblige, but it does not have my preference.

ederag:
Boy. How could that be OK ?
Just take my commits, as is, like I [asked](https://github.com/projecthamster/hamster/pull/630#issuecomment-798115783).
As if I had committed those before leaving.
Responsibility is mine, if that appease you.

matthijskooijman:
> Just take my commits, as is, like I asked.

As I explained, I do not think the commit messages are verbose enough and some commits need minor restructuring, so this is not an acceptable option for me.

I'm going to assume this means you'd rather have new commits without you listed as the author, than modified commits with you listed as the author, so I'll do that, then.

ederag:
> you'd rather have new commits without you listed as the author, than modified commits with you listed as the author, so I'll do that, then.

No I explicitly **deny** you that right.

matthijskooijman:
> After the hundreds of hours I spent on this project, a minimum of respect would be expected.

I'm trying my best to do this in a respectful and open manner, additionally discussing the approach I want to take and looking for options that would be acceptable for everyone. I get the impression that you consider your own commits and commit messages being fully perfect and that you consider that *any* change to them makes them worse, even without even knowing exactly what changes those would be. I'm sorry you feel disrespected, that is certainly not my intention.

> No I explicitly deny you that right.

Ok, I've noted your intent and I'll see what I can do without copying any of your code. For the record, I do not think your denial is legally feasible (the GPL grants an irrevocable right), but I'll try to respect it regardless (but I expect some bits of code might end up identically anyway).

ederag:
You are all so mistaken about how free software should work.

> I'm trying my best to do this in a respectful and open manner

Respect is much more than civility.
And respect there's little in this whole thread.

rhertzog:
> No I explicitly deny you that right.

You can't do that. The license explicitly permits anyone from reusing your code. I'm not sure why you are still here... if you don't want people to reuse your code and you only contribute poisonous remarks, then there's no reason for us to tolerate your presence here, whatever historical contribution you may have made.

Issue #587, opened 2020-03-17
Improve and automate testing


matthijskooijman:
It would be good if testing the code could be automated, in the sense that all pullrequests and merges to master are automatically tested by something like Travis. This makes it easier to spot bugs.

Additionally, some style enforcement (pep8 checking e.g. @ederag mentioned planning to use black in https://github.com/projecthamster/hamster/issues/586#issuecomment-599664483) would be good, but that might require quite some investment to make the code compliant first.

Finally, the code currently uses the `unittest` framework, @ederag suggested he was planning to move to `pytest` instead. I haven't compared both frameworks, I guess pytest is more powerful or easy to work with maybe?

mwilck:
If we want better CI, the primary todo item would be writing more tests, not switching to new framework.

matthijskooijman:
I had a look at this today, it seems that Github actions would be an easy way to get automated checks, for example on all pull requests and the latest master branch.

Running unit tests is probably a good start.

I would also like to run some code-style (e.g. PEP-8/flake8) and linting checks. One challenge there is that the current codebase does not seem to comply with PEP-8, and running a linter over it will probably produce a lot of warnings already. Ideally, we would fix the code to be clean, and then enforce it to stay clean with CI checks after that.

Since fixing all code will be a lot of work (running `flake8` on the current master, excluding `waflib`, already produces 2089 warnings...), we might want to automate this.

For code-style, there is a relatively new tool called `black`, which automates formatting of Python code. It is intentionally not very configurable and somewhat opiniated, which I would think is a fair deal given it just results in a very consistent style (also between other projects that use black) and you can just code and run `black` afterwards to fix any inconsistencies. Black essentially applies a stricter version of PEP8 (PEP8 leaves some things open for choice, where Black makes a single choice).

I'll create a draft PR soon, so we can at least see what changes this would result in.

A downside of a lot of (automated) changes is of course that git blame results will be less usable and porting commits between branches might get more tricky...

I'm not sure if there are any automated lint-checker fixers that can do a similar thing, so linting is probably going to be harder.

mwilck:
> A downside of a lot of (automated) changes is of course that git blame results will be less usable and porting commits between branches might get more tricky...

Indeed. I have little experience with python linting, but I'd recommend to avoid whitespace changes as much as possible.

matthijskooijman:
[...]

PR #600, opened 2020-04-29
Reformat with black

matthijskooijman:
This is an experimental PR, just to see what a reformat with [the black formatter](https://github.com/psf/black/) would look like. See https://github.com/projecthamster/hamster/issues/587#issuecomment-621159567 for previous discussion.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,13 +6,14 @@ from distutils import sysconfig
 
 data_dir = os.path.join(sysconfig.get_python_lib(), "hamster", "data")
 
-setup(name='hamster-sqlite',
-      version='0.3',
-      description='Minimal dependency nicely abstracted sqlite backend of hamster time tracker - lets you connect to your hamster db and do stuff in python',
-      author='Toms Baugis',
-      author_email='toms.baugis@gmail.com',
-      url='https://github.com/projecthamster/hamster',
-      package_dir = {'': 'src'},
-      py_modules = ['hamster.storage', 'hamster.db', 'hamster.lib.__init__'],
-      data_files=[(data_dir, ['data/hamster.db'])],
-     )
+setup(
+    name='hamster-sqlite',
+    version='0.3',
+    description='Minimal dependency nicely abstracted sqlite backend of hamster time tracker - lets you connect to your hamster db and do stuff in python',
+    author='Toms Baugis',
+    author_email='toms.baugis@gmail.com',
+    url='https://github.com/projecthamster/hamster',
+    package_dir={'': 'src'},
+    py_modules=['hamster.storage', 'hamster.db', 'hamster.lib.__init__'],
+    data_files=[(data_dir, ['data/hamster.db'])],
+)
--- a/setup.py
+++ b/setup.py
@@ -7,13 +7,13 @@ from distutils import sysconfig
 data_dir = os.path.join(sysconfig.get_python_lib(), "hamster", "data")
 
 setup(
-    name='hamster-sqlite',
-    version='0.3',
-    description='Minimal dependency nicely abstracted sqlite backend of hamster time tracker - lets you connect to your hamster db and do stuff in python',
-    author='Toms Baugis',
-    author_email='toms.baugis@gmail.com',
-    url='https://github.com/projecthamster/hamster',
-    package_dir={'': 'src'},
-    py_modules=['hamster.storage', 'hamster.db', 'hamster.lib.__init__'],
-    data_files=[(data_dir, ['data/hamster.db'])],
+    name="hamster-sqlite",
+    version="0.3",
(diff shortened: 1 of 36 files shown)


setup.py
```diff
@@ -7,13 +7,13 @@
...
-    name='hamster-sqlite',
-    version='0.3',
-    description='Minimal dependency nicely abstracted sqlite backend of hamster time tracker - lets you connect to your hamster db and do stuff in python',
-    author='Toms Baugis',
-    author_email='toms.baugis@gmail.com',
-    url='https://github.com/projecthamster/hamster',
-    package_dir={'': 'src'},
-    py_modules=['hamster.storage', 'hamster.db', 'hamster.lib.__init__'],
-    data_files=[(data_dir, ['data/hamster.db'])],
+    name="hamster-sqlite",
```
mwilck, line 10: What's the point of replacing `'` with `"`? I thought both are equivalent in python, is that no longer true?

setup.py
```diff
@@ -6,13 +6,14 @@
...
-      version='0.3',
-      description='Minimal dependency nicely abstracted sqlite backend of hamster time tracker - lets you connect to your hamster db and do stuff in python',
-      author='Toms Baugis',
-      author_email='toms.baugis@gmail.com',
-      url='https://github.com/projecthamster/hamster',
-      package_dir = {'': 'src'},
-      py_modules = ['hamster.storage', 'hamster.db', 'hamster.lib.__init__'],
-      data_files=[(data_dir, ['data/hamster.db'])],
-     )
+setup(
```
mwilck, line 9: Much better without the quote replacement, but I see little value in this indentation change either.

src/hamster-cli.py
```diff
@@ -109,10 +110,12 @@ class Hamster(gtk.Application):
     def __init__(self):
         # inactivity_timeout: How long (ms) the service should stay alive
         #                     after all windows have been closed.
-        gtk.Application.__init__(self,
-                                 application_id="org.gnome.Hamster.GUI",
-                                 #inactivity_timeout=10000,
-                                 register_session=True)
+        gtk.Application.__init__(
```
mwilck, line 113: Same as above, doesn't improve readability, doesn't even avoid over-long lines.

src/hamster-cli.py
```diff
@@ -277,8 +276,9 @@ def export(self, *args):
         end_time = end_time or start_time.replace(hour=23, minute=59, second=59)
         facts = self.storage.get_facts(start_time, end_time)
 
-        writer = reports.simple(facts, start_time.date(), end_time.date(), export_format)
-
+        writer = reports.simple(
```
mwilck, line 279: Ugh! That's a deterioration IMO

src/hamster-cli.py
```diff
@@ -314,17 +312,18 @@ def list(self, *times):
...
         self._list(start_time, end_time)
 
-
     def current(self, *args):
         """prints current activity. kinda minimal right now"""
         facts = self.storage.get_todays_facts()
         if facts and not facts[-1].end_time:
-            print("{} {}".format(str(facts[-1]).strip(),
-                                 facts[-1].delta.format(fmt="HH:MM")))
+            print(
```
mwilck, line 319: Again, rather worse than better as before to my taste

src/hamster-cli.py
matthijskooijman: This is just because the line is > 88 characters, so it is wrapped. Might mean the line length should be made a bit longer, then.

matthijskooijman: I just looked at the current code to see if there's a line length that would minimize the diff, but it seems the current code is not all that consistent, so that probably won't really work.

matthijskooijman:
@mwilck, I suggest you look through the black README, there they explain a bunch of the choices they made (such as the quotes): https://github.com/psf/black

Overall, I suspect that in a lot of cases the changes made in this PR are not necessarily better than what was there, the gain is mostly that things are consistent and formatting can therefore be applied automatically without having to think much.

mwilck:
@matthijskooijman, I have no doubts that they have a rationale for what they do. Others will have similarly good arguments for their different style preferences. At the end of the day the project, i.e. **us**, must figure out what we want.

For me, `git blame` is the killer argument against changes like this. When reading or debugging  code, it's very often necessary to use `blame` or similar tools to find out why things were done the way they are. Broken `blame` makes that unnecessarily hard. 

I'd favour an approach like the kernel's `checkpatch.pl`, where style checking is only applied to the *lines added* by a given patch. IOW, be strict with new code, but be forgiving with the existing code base. When a patch changes a bunch of old code for non-style reasons, it can then fix the style along the way, too.

matthijskooijman:
As it stands now, I think black is not the tool to use, at least not right now, so I'm going to close this PR.

I've created a new issue, #606, to track the more general issue there. It also has some ideas about a new git blame feature that could be helpful here, so let's continue discussion there.

PR #601, opened 2020-04-30
Support python 3.5

matthijskooijman:
This PR makes some small changes to support Python3.5 again. It should not change behaviour in any way.

PR #602, opened 2020-04-30
Run flake8 and unittests on pullrequests and pushes

matthijskooijman:
This sets up Github Actions to automatically run flake8 and unittests on pullrequests and direct pushes. Test results should show up below the pullrequest and the author might get an e-mail notification (at least for direct pushes to this repository, not sure about pullrequests).

There are some additional commits to simplify running the test suite and document this in the README.

This PR assumes that the minimum supported Python version is 3.5, so it runs tests and flake8 under 3.5. Since this assumption is not currently true, these tests will now fail until #601 is merged. The testsuite is also ran under a newer Python version, which should already succeed now.

Since the code is not currently pep8-compliant, flake8 is run with a *lot* of ignores set, to allow it to complete succesfully (after #601 is merged). This is not ideal, but until we fix these pep8 issues, this check will at least ensure that no new kinds of violations will be introduced.


README.md
```diff
@@ -151,6 +151,24 @@ D-Bus, so that all the traces are visible.
 Note: You'll need recent version of hamster installed on your system (or 
 [this workaround](https://github.com/projecthamster/hamster/issues/552#issuecomment-585166000)).
 
+#### Running tests
+
+Hamster has a limited test suit, that can be run using Python's builtin
```
mwilck, line 156: suit -> suite

matthijskooijman:
> Would it choke on the current code?

Normally, yes. In this case, I configured flake8 to ignore a lot of (types of) violations, so it can at least run successfully. It will not check as much as it could, but we can later see which things we would like to fix or not and remove the ignores again. There is a tool called autopep8 which is similar to black, but more configurable, so we could maybe use that to more selectively fix things and keep ignoring others).

> Or can we adjust it such that only new code additions would be rejected if they don't conform?

I'm not sure, that is probably tricky (you'd have to figure out what changed, which is a bit more complicated, then somehow filter the results, which I think flake8 itself does not support, and there's always the issue that moving code around without modifying it, or making unrelated changes, would then force you to change the formatting, which likely results in much more commits, or commits mixing different kinds of changes).

In any case, all of this, I think, is out of scope for this PR and food for later improvements.

mwilck:
> In any case, all of this, I think, is out of scope for this PR and food for later improvements.

I agree.

Currently the CI seems to be choking on the problems you pointed out in #601. You have my blessing to merge that quickly (I think it's obvious enough to be merged with two mainainers - you and me - approving). I'd like to see the CI succeed after that merge before I approve this PR. A CI that always fails has no value; on the contrary, it scares people away, users as well as contributors and maintainers.

matthijskooijman:
> Currently the CI seems to be choking on the problems you pointed out in #601. You have my blessing to merge that quickly

Yeah, I'll do that, and also fix the typo you pointed out :-)

matthijskooijman:
I rebased on top of master (which has py3.5 support from #601 merged), which makes the tests run succesfully. In addition, I:
 - Fixed the typo pointed out by @mwilck 
 - Added one more dependency to the README (turns out the Ubuntu VM had `libglib2.0-bin` installed by default, I noticed this on building a .deb in an even more bare chroot).
 - Reworked the way dependency differences between Ubuntu 16 and 18 were done (previously I used regex to just ignore any missing packages, now it just checks the OS version, which is more explicit and cannot accidentally mask problematic missing packages).

I left these in fixup commits for easier review, I'll need to squash those before this PR can be merged.

matthijskooijman:
Ok, just squashed the fixup commits, and with three approvals, I'll merge this.

PR #610, opened 2020-05-18
Add a flatpak manifest

tchernobog:
This PR adds flatpak support for Hamster. Building the application is surprisingly easy, and making it available on flathub should contribute to Hamster's usage growth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,33 @@ Installation:
 Easy installation on any distribution supporting snap:  
 https://snapcraft.io/hamster-snap
 
+##### Flatpak
+
+[Flatpak](https://flatpak.org/) enables you to install Hamster in a versioned
+environment and then run it inside a sandbox. It is a method independent from
+your distribution-specific packaging system, ensuring that the application can
+always be reproducibly built, even without hunting down all of the dependencies
+yourself. Debugging is made easier as every user has the exact same environment
+at runtime. Permissions are limited to what the application really needs to
+function properly.
+
+If you downloaded the file with the Hamster bundle (ending in ``.flatpak``), you
+can directly install it with:
+
+```bash
+flatpak install --reinstall Hamster.flatpak
+```
+
+If you would like to install Hamster only for your user, please pass the
+``--user`` option to the above command.
+
+After installation, if you need to invoke Hamster from the command line,
+you can do so with:
+
+```bash
+flatpak run org.gnome.Hamster [args...]
+```
+
 ### Install from sources
 
 #### Dependencies
@@ -125,6 +152,20 @@ as discussed [here](https://github.com/projecthamster/hamster/pull/421#issuecomm
 
 Now restart your panels/docks and you should be able to add Hamster!
 
+##### Flatpak
+
+Alternatively, you can also build a sandboxed
+[flatpak](https://www.flatpak.org/) yourself. You might need to install the
+GNOME SDK beforehand (an error will notify you about it, if needed). Execute:
+
+```bash
+flatpak-builder --force-clean --user --install \
+    build/flatpak org.gnome.Hamster.json
+```
+
+This creates a temporary flatpack build folder in the ``build/flatpak``
+directory. Once the app is installed, the whole ``build/flatpack/`` directory
+can be removed.
 
 #### Uninstall
 
@@ -134,6 +175,14 @@ sudo ./waf uninstall
 ```
 Afterwards `find /usr -iname hamster` should only list unrelated files (if any).
 Otherwise, please see the [wiki section](https://github.com/projecthamster/hamster/wiki/Tips-and-Tricks#uninstall)
+
+##### Flatpak
+
+To remove the installed flatpak, just run:
+
+```bash
+flatpak uninstall org.gnome.Hamster
+```
 
 #### Troubleshooting
 
@@ -178,7 +227,31 @@ run:
     python3 -m unittest tests.test_stuff.TestFactParsing
     python3 -m unittest tests.test_stuff.TestFactParsing.test_plain_name
 
-#### Migrating from hamster-applet
+##### Flatpak
+
+To run the tests inside the flatpak, use:
+
+```bash
+flatpak-builder --run build/flatpak org.gnome.Hamster.json \
+    python3 -m unittest
+```
+
+#### Migrating
+
+##### Migrating data to flatpak
+
+If you would like to retain your data from a non-flatpak installation,
+you can do so by running:
+
+```bash
+gio copy -b \
+    ~/.local/share/hamster/hamster.db \
+    ~/.var/app/org.gnome.Hamster/data/hamster/
+```
+
+After checking everything works, you can remove the original database.
+
+##### Migrating from hamster-applet
 
 Previously Hamster was installed everywhere under `hamster-applet`. As
 the applet is long gone, the paths and file names have changed to
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -77,3 +77,42 @@ jobs:
       - name: Run flake8
         run: |
           flake8 --count --show-source --statistics
+
+  flatpak:
+    # This job builds Hamster for flatpak, and runs its tests
+    # inside the sandbox.
+    name: Test the flatpak build
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Prepare repo
+        uses: actions/checkout@v2
+      - name: Install system packages
+        # It would be good to cache the GNOME Sdk, as it
+        # is rather big to download each time.
+        env:
+          DEBIAN_FRONTEND: noninteractive
+        run: |
+          PACKAGES="flatpak flatpak-builder dbus-x11"
+
+          sudo apt-get update
+          sudo apt-get install -yq ${PACKAGES}
+      - name: Install GNOME SDK for flatpak
+        run: |
+          flatpak remote-add --user --if-not-exists --from flathub https://flathub.org/repo/flathub.flatpakrepo
+          flatpak install --user -y flathub org.gnome.Platform//3.36 org.gnome.Sdk//3.36
+      - name: Build application
+        run: |
+          flatpak-builder --repo=build/flatpak/repo build/flatpak/tmp org.gnome.Hamster.json
+      - name: Run tests inside sandbox
+        run: |
+          dbus-launch flatpak-builder --run build/flatpak/tmp org.gnome.Hamster.json python3 -m unittest
+      - name: Export bundle and try to install it
+        run: |
+          mkdir -p dist
+          flatpak build-bundle --runtime-repo=https://flathub.org/repo/flathub.flatpakrepo build/flatpak/repo dist/Hamster.flatpak org.gnome.Hamster
+          flatpak --user -y install dist/Hamster.flatpak
+      - name: Upload built artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: Flatpak application
+          path: dist/Hamster.flatpak
--- a/org.gnome.Hamster.json
+++ b/org.gnome.Hamster.json
@@ -1,0 +1,68 @@
+{
+    "app-id": "org.gnome.Hamster",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "3.36",
+    "sdk": "org.gnome.Sdk",
+    "command": "hamster",
+    "modules": [
+        {
+            "name": "intltool",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://launchpad.net/intltool/trunk/0.51.0/+download/intltool-0.51.0.tar.gz",
+                    "md5": "12e517cac2b57a0121cda351570f1e63"
+                }
+            ],
+            "cleanup": [ "*" ]
+        },
+        {
+            "name": "python3-hamster-dependencies",
+            "buildsystem": "simple",
+            "ensure-writable": [
+                "easy-install.pth",
+                "setuptools.pth"
+            ],
+            "build-commands": [
+                  "pip3 install --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} dbus-python pyxdg"
+            ],
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/62/7e/d4fb56a1695fa65da0c8d3071855fa5408447b913c58c01933c2f81a269a/dbus-python-1.2.16.tar.gz",
+                    "sha256": "11238f1d86c995d8aed2e22f04a1e3779f0d70e587caffeab4857f3c662ed5a4"
+                },
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/39/03/12eb9062f43adb94e30f366743cb5c83fd15fef026500cd4de42c7c12280/pyxdg-0.26-py2.py3-none-any.whl",
+                    "sha256": "1948ff8e2db02156c0cccd2529b43c0cff56ebaa71f5f021bbd755bc1419190e"
+                }
+            ]
+        },
+        {
+            "name": "hamster",
+            "buildsystem": "simple",
+            "builddir": true,
+            "prefix": "/app",
+            "build-commands": [
+                "./waf configure --prefix=${FLATPAK_DEST}",
+                "./waf build",
+                "./waf install"
+            ],
+            "sources": [
+                {
+                        "type": "dir",
+                        "path": "."
+                }
+            ]
+        }
+    ],
+    "finish-args": [
+        "--socket=wayland",
+        "--socket=fallback-x11",
+        "--filesystem=xdg-documents",
+        "--own-name=org.gnome.Hamster",
+        "--own-name=org.gnome.Hamster.GUI",
+        "--own-name=org.gnome.Hamster.WindowServer"
+    ]
+}


mwilck:
This is great! I've thought about doing this myself, thanks a lot.

Have you tried if this works with the GNOME shell extension? See [hamster-shell-extension#328](https://github.com/projecthamster/hamster-shell-extension/issues/328) for the issues we had with the snap package (permission problems with DBus service autostart). It'd be great news if flatpak didn't have these issues.

tchernobog:
@mwilck Yes, I am actually running it with the GNOME shell extension, and was pleasantly surprised to see it works out of the box!

src/hamster/lib/configuration.py
```diff
@@ -26,7 +26,7 @@
 
 import os
 from hamster.client import Storage
-from xdg.BaseDirectory import xdg_data_home
+from xdg import XDG_DATA_HOME
```
mwilck: This code fails on my system:

    Python 3.8.2 (default, Mar 05 2020, 18:58:42) [GCC] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> from xdg import XDG_DATA_HOME
      File "<stdin>", line 1, in <module>
    ImportError: cannot import name 'XDG_DATA_HOME' from 'xdg' (/usr/lib/python3.8/site-packages/xdg/__init__.py)


OTOH, `from xdg.BaseDirectory import xdg_data_home` works for me. It seems to be yet another python incompatibility. So we'll need to do something like

    try:
        from xdg.BaseDirectory import xdg_data_home   
    except ImportError:
        try:
            from xdg import XDG_DATA_HOME as xdg_data_home
        except ImportError:
            xdg_data_home = os.environ.get('XDG_DATA_HOME')
    if xdg_data_home is None:
        xdg_data_home="{}/.local/share".format(os.environ.get("HOME"))

Not beautiful, but apparently necessary.

tchernobog: I reverted that in the latest commit, thanks for pointing that out.

matthijskooijman: Looks like this is a matter of depending on [`xdg`](https://pypi.org/project/xdg/) or [`pyxdg`](https://pypi.org/project/pyxdg/), which are two different packages for accessing xdg (and also both exporting the `xdg` top-level python package, making them incompatible I think). This PR would switch from one to the other, which @mwilck probably did not have installed.

Switching from one to the other could be done, but is probably out of scope for this PR, so good to see that @tchernobog updated the PR to just install `pyxdg` instead.

Given these two conflict, I guess that supporting both with the code @mwilck might be useful too (in case someone has `xdg` installed for another application and thus cannot install `pyxdg`), but that's also for a different PR (and mabye we should not bother until someone actually needs this). Leaving `pyxdg` as the default is probably a good idea in any case, since that seems to be hosted and/or maintained by freedesktop.org and it is also shipped by Debian (under the somewhat confusing name `python3-xdg`, which I think refers to the top-level package rather than the upstream project name)...

matthijskooijman:
Just quickly read up on flatpack, sounds like an interesting way to distribute hamster directly to users (in addition to through distro packages), probably better than letting users run `waf install` directly, with all the potentially untracked mess on a system resulting from that.

I am wondering what the implications of adding this are. How would things work, what would needed be done when?

Some specific questions:
 - The flatpack manifest contains explicitly versioned dependencies. I guess we would need to manually bump those versions every now and then. Is there tooling to find the latest version, or would this mean manually updating version numbers and checksums?
 - It seems this PR only facilitates a local build of flatpack, which is a great start. But flatpack is intended to used with repositories and auto-updating, is that something we should think about in the longer term? Would that mean uploading to some central flatpack repo? Or would we need to host our own?
 - It seems useful to automatically build a flatpacks for each pull request (for easier testing) and maybe releases too. Does flatpack support packaging into a single downloadable file, or would that require going through a repository of sorts?
 - The additions to the README in this PR are a little concise. I would welcome some more additions, so a user that has never really worked with flatpack knows a bit better what they are doing (e.g that a flatpack repo is created, into which a flatpack application is built, which is later installed from the repo).
 - The manifest specifies the gnome 3.36 runtime. Does this support running on other gnome versions too? I suspect it does, since hamster itself does not depend on Gnome anymore, I think, it only needs GTK and some build tools and maybe icons (and I guess the gnome runtime is the easiest way to install all that?). Only the extension depends on gnome(-shell) more closely, but that is installed normally outside of flatpack, right?

tchernobog:
> Some specific questions:
> 
>     * The flatpack manifest contains explicitly versioned dependencies. I guess we would need to manually bump those versions every now and then. Is there tooling to find the latest version, or would this mean manually updating version numbers and checksums?

Usually it's done manually as part of the packaging effort. It would be possible to extract this somewhat automatically if Hamster were to use Pipenv and a Pipfile.lock, else I personally don't think it's worth the effort.

>     * It seems this PR only facilitates a local build of flatpack, which is a great start. But flatpack is intended to used with repositories and auto-updating, is that something we should think about in the longer term? Would that mean uploading to some central flatpack repo? Or would we need to host our own?

I wanted first to enable people to build a flatpak locally, which is great for testing. Publishing on flathub.org is easy enough, but then the manifest would need to be moved out of this repository and into flathub's. See https://github.com/flathub/flathub/wiki/App-Submission#how-to-submit-an-app

>     * It seems useful to automatically build a flatpacks for each pull request (for easier testing) and maybe releases too. Does flatpack support packaging into a single downloadable file, or would that require going through a repository of sorts?

That would be a single-file bundle: https://docs.flatpak.org/en/latest/single-file-bundles.html
It can be done as part of continuous integration.

> 
>     * The additions to the README in this PR are a little concise. I would welcome some more additions, so a user that has never really worked with flatpack knows a bit better what they are doing (e.g that a flatpack repo is created, into which a flatpack application is built, which is later installed from the repo).

This PR is a bit more targeted at developers wanting to test it out (so one assumes a bit of knowledge of flatpak anyway), but sure, I can add that.

>     * The manifest specifies the gnome 3.36 runtime. Does this support running on other gnome versions too? I suspect it does, since hamster itself does not depend on Gnome anymore, I think, it only needs GTK and some build tools and maybe icons (and I guess the gnome runtime is the easiest way to install all that?). Only the extension depends on gnome(-shell) more closely, but that is installed normally outside of flatpack, right?

If Hamster is GNOME-agnostic, we can depend on the Freedesktop runtime. But the main idea is that flatpak forces you to decide which set of libraries (here, GNOME 3.36 libraries) to target, so that each system installing your app will have exactly the same set of code running. That means ease of debugging, a lot less unreproducible bugs, etc.

tchernobog:
I addressed the comments by adding some more documentation in the README, and showing how to build a bundle instead of a repository.

mwilck:
> If Hamster is GNOME-agnostic, we can depend on the Freedesktop runtime.

I'm unaware of any GNOME version dependencies of hamster itself. The GNOME extension has more than enough though.

matthijskooijman:
> I also tried to move away from the GNOME SDK and use the Freedesktop SDK directly, but some dependencies are not available (for instance, gobject-introspection support).

Yeah, we don't rely on Gnome itself (that is, you can run hamster on KDE just fine, I think), but we *do* need some gnome-related libraries for the UI (gtk/glib/gobject), so it is probably easier to just depend on (any version of) the gnome runtime than use the freedesktop runtime and install all the needed libraries on top of that.

> This PR is a bit more targeted at developers wanting to test it out (so one assumes a bit of knowledge of flatpak anyway), but sure, I can add that.

Yeah, looks much better (though it also makes a lot of unrelated whitespace changes, could you revert those?). Going through a bundle instead of installing from a local repo is nice because:
 - The installation step is identical between self-building and building from a CI build (which makes instructions simpoler).
 - There is no longer any "one-time-setup" command for adding the repo
 - The repo is now just basically a temporary build directory.

 The last point could maybe be made explicit in the README? Something like:

> This creates a temporary flatpack repository in the `flatpack/repo` directory containing the built application, which is then exported to a `.flatpack` file. The `.flatpack` file can be installed as above, the `flatpack/repo` directory can be removed.

I also see a `flatpack/build` argument, is that actually "build" command being ran? Or is that a temporary build directory? If the latter, it should be included in the previous note as well, I think.



One worry I have is that if we merge this, we also need to make sure the manifest file stays working. Some ingredients to tackle that:
 - Building the flatpack bundle as a github action would at least ensure that it stays buildable, and makes it easy for anyone to test the flatpack build.
 - Some manual testing of the build before each release would be useful. This would work best by people that already run the flatpack build normally, I guess. @tchernobog, would you yourself in such a role?
 - More extensive automated testing (e.g. running the testsuite inside the flatpack build, or maybe just some tests that talk to the dbus interface) could also help, but I'm not sure where to begin with that.

matthijskooijman:
Hm, I just noticed #561, which also intends to add a flatpack manifest. @bochecha, you created #561, can I invite you to participate in this PR as well?

Overall, I'm inclined to work towards merging this PR over #561, since this PR adds some docs to the README. I also do not quite like the `build-aux` top level folder name (though I think I do like having some kind of top-level folder, to collect build-related stuff, rather than ending up with `snap`, `flat` and more directories like that, but maybe we should just separately review the folder structure later).

I also compared the flatpack manifests themselves a bit, and it seems they are mostly identical, but with some small differences. @tchernobog, @bochecha, since you are more familiar with flatpak, maybe you could compare both so we can pick the best from both?

One thing I did notice in the description of #561 is that `flatpackbuilder` apparently has a `--install` option, which I think builds and installs directly. If I got that correctly, then that might be even nicer to use as a "build yourself" command in the README?

tchernobog:
> One worry I have is that if we merge this, we also need to make sure the manifest file stays working. Some ingredients to tackle that:

>    Building the flatpack bundle as a github action would at least ensure that it stays buildable, and makes it easy for anyone to test the flatpack build.

Yes, I agree.

> Some manual testing of the build before each release would be useful. This would work best by people that already run the flatpack build normally, I guess. @tchernobog, would you yourself in such a role?

I can do that if there is a way for me to be notified a release is coming.

> More extensive automated testing (e.g. running the testsuite inside the flatpack build, or maybe just some tests that talk to the dbus interface) could also help, but I'm not sure where to begin with that.

It is possible to get a shell inside the sandboxed build environment and run tests with:
```flatpak-builder --run build/flatpak/tmp org.gnome.Hamster.json python3 -m unittest```
after invoking flatpak-builder as per README.md.

I reviewed the awesome work of @bochecha, which for some reason I missed before (sorry!). There are not a lot of differences; mostly the owned D-Bus names, relying to an external git submodule for intltools (but I guess Hamster should move to plain gettext as intltool is deprecated, and it's not a lot of code), and removing the icon cache.

@bochecha Can you elaborate on the need to remove the icon cache? It's the only item I am unsure about. Thanks!

**Comment by matthijskooijman** [comment]
>  (but I guess Hamster should move to plain gettext as intltool is deprecated, and it's not a lot of code)

I think that we're not actually using `intltool` itself anymore, just the `intltool` waf module to call `gettext`. Maybe some refactoring on the waf side could be done to drop that dependency, haven't looked at it more closely yet.

**Comment by tchernobog** [comment]
I added a pipeline for building and testing the flatpak bundle.

**Comment by benjaoming** [comment]
From description:
> making it available on flathub should contribute to Hamster's usage growth.

Later:
> But the main idea is that flatpak forces you to decide which set of libraries (here, GNOME 3.36 libraries) to target, so that each system installing your app will have exactly the same set of code running. That means ease of debugging, a lot less unreproducible bugs, etc.

I immediately got confused about the intention of having Flatpak here.. Should we encourage a user of Gnome 3.34 to install a Flatpak version of this small extension with Gnome 3.36 libraries in a Flatpak container? And if they already have Gnome 3.36 in their system-wide runtime, will this Flatpak add a second version of the same libraries (I don't know so much about Flatpak, so that's an honest question)? These are speculative questions, but my pending conclusion would be: That's a lot of overhead for a small extension.

For simplifying installation, I would go with the `gnome-shell-extension-tool` utility that I just discovered can be a quick and easy shortcut to install. And then of course to get a new hamster website up and running :)

I also think we should be vary that we already had the problem of 3 different versions on extensions.gnome.org and Flathub could introduce the same problem.

We can open up a separate issue about user-oriented Flatpak distribution -- for now, would be great to keep the effort for Debian/Ubuntu packaging.

When it comes to having Flatpak for testing-purposes, it seems like a great shortcut, the implementation here is quick. Would also be great to have a test matrix so we can target the tests to supported Gnome versions.

**Comment by tchernobog** [comment]
@matthijskooijman, @mwilck: are there any other changes you would like? Is there still something you need before approval can be given?

**Comment by tchernobog** [comment]
> One new question I realized: How are datafiles handled (e.g. if you upgrade or remove the snap, where does the database go)? In the manifest I see `--filesystem=xdg-documents`, I suspect that means that e.g. `~/.local` and `~/.config` from the host system are made available inside the flatpack?

Not a snap, it's a flatpak :-D. It creates its own sandboxed folder for the application. Data files are kept in `$HOME/.var/org.gnome.Hamster/`. These are persisted across upgrades or uninstall, just like a non-flatpak app.

> As for the the PR itself, the history is a bit messy right now, we should probably squash some commits before merging. 

I will do this myself as the last step, after a final review of the result from you.

**Comment by matthijskooijman** [comment]
> Data files are kept in $HOME/.var/org.gnome.Hamster/

Ah, so that means they are different from the datafiles used by a normal install. I wonder if it is possible to just use the same datafiles as a "normal" install (e.g. allow access to `~/.local` and `~/.config` I think)? And if it is, we should decide whether we want this. At first glance, it seems sensible, to get the cleanest upgrade path from / to the flatpack install.

@tchernobog @mwilck what would you think?

> I will do this myself as the last step, after a final review of the result from you.

Ok, great.

**Comment by tchernobog** [comment]
> that means they are different from the datafiles used by a normal install. 
>
> @tchernobog @mwilck what would you think?
>

I personally would prefer to avoid this. It would break the concept of sandboxing. Also, the user might want to keep installed a different system version of Hamster (via package manager) and the new one might be incompatible. If you use flatpak, it's normally accepted that applications have restricted access to your $HOME (including previous settings).

**Comment by matthijskooijman** [comment]
> I personally would prefer to avoid this. It would break the concept of sandboxing. Also, the user might want to keep installed a different system version of Hamster (via package manager) and the new one might be incompatible.

Yeah, that's a good point. Might be good to document this, then, and maybe offer some advice on what files need to be copied to migrate data into and out of flatpack. I guess that migrating configuration might be more tricky (since that lives in dconf I think), but that's probably ok, I guess moving data is more relevant anyway.

**Comment by tchernobog** [comment]
> I guess that migrating configuration might be more tricky (since that lives in dconf I think), but that's probably ok, I guess moving data is more relevant anyway.

I added a section on migrating the database. For DConf, supposedly there would be a way to trigger a migration of config: https://docs.flatpak.org/en/latest/sandbox-permissions.html#dconf-access

However, I tried it and it seems it breaks running the unit tests for some reason, so I did not include it as of this time.

**Comment by tchernobog** [comment]
@matthijskooijman are there any pending changes you would like me to perform before merging?

**Comment by matthijskooijman** [comment]
@tchernobog not off-hand, I still want to do a review of the whole and your new additions, but no time right now (finishing up some other things before my vacation, so I'll probably won't get around to this until after my vacation, so early august).

Also, would be good if one of the other maintainers could do a full review as well. Maybe @mwilck? You have looked at this early on already?

**Comment by tchernobog** [comment]
@mwilck would it be possible to merge this?

**Comment by mwilck** [comment]
@matthijskooijman, AFAICS the change you'd requested previously has been addressed.

**Comment by magicmyth** [comment]
Just a fly by comment. I just installed the new hamster via this flatpak and it worked fine on my KDE Neon 18.04 machine. I followed the instructions in the README.md and copied over my existing database (from 2.91 Ubuntu package) and was able to get right back to work. I'm no expert but I've made a few flatpak builder recipies and tchernobog's work looks perfectly fine to me.

One sugestion I have is to maybe add to the docs about needing to run cli commands as `flatpak run org.gnome.Hamster [CMD]` instead of just `hamster [CMD[`. Might be worth giving instructions on how to alias that for convenience  E.g `alias hamster='flatpak run org.gnome.Hamster'`.

**Comment by mwilck** [comment]
> One sugestion I have is to maybe add to the docs about needing to run cli commands as `flatpak run org.gnome.Hamster [CMD]` instead of just `hamster [CMD[`. Might be worth giving instructions on how to alias that for convenience E.g `alias hamster='flatpak run org.gnome.Hamster'`.

Thanks. @tchernobog, could you add that?

**Comment by tchernobog** [comment]
> > One sugestion I have is to maybe add to the docs about needing to run cli commands as `flatpak run org.gnome.Hamster [CMD]`

> Thanks. @tchernobog, could you add that?

Done, I would prefer to proceed with a merge however now before performing further changes. Making it perfect can be left to subsequent commits.

**Comment by mwilck** [comment]
@tchernobog, thanks a lot.

> I would prefer to proceed with a merge

Sure. As you can see, @matthijskooijman's review is still pending. I trust that as soon as he gets to it, he'll see my positive review and proceed with the merge.

File: README.md [hunk 2]
```diff
@@ -125,6 +152,20 @@ as discussed [here](https://github.com/projecthamster/hamster/pull/421#issuecomm
 
 Now restart your panels/docks and you should be able to add Hamster!
 
+##### Flatpak
+
+Alternatively, you can also build a sandboxed
```
matthijskooijman, line 157: This suggests it just builds the flatpack, but I think it also installs it. Might be good to clarify.

README.md
```diff
@@ -58,6 +58,33 @@ Installation:
...
+```
+
+If you would like to install Hamster only for your user, please pass the
+``--user`` option to the above command.
+
+After installation, if you need to invoke Hamster from the command line,
+you can do so with:
+
+```bash
+flatpak run org.gnome.Hamster [args...]
```
matthijskooijman, line 85: Would there be any easy way to make this shorter? Maybe flatpack can automically put a symlink in `~/bin` or `/usr/local/bin`? Or otherwise register a short name so you can at least use `flatpack run hamster` (i.e. using the same base command as a normal install)?

Also, it might be good to clarify that (IIUC) the flatpack-hamster will be autostarted through dbus (e.g. when the shell extension tries to access it). That does make me wonder: What does flatpack do when the dbus name is already registered (i.e. when hamster is normally installed as well)? From a quick test it seems to (silently) not claim the name, that is, when I start the flatpack, the flatpack GUI connects to the non-flatpack dbus service, and when I kill the non-flatpack services, it still spawns the non-flatpack ones, rather than the flatpack ones.

For most users, I guess this is not so much an issue (they would typically only install the flatpack version), but it would be at least to document these caveats somewhat. Also, one usecase I see for flatpack is that it would help with testing nightly builds or pullrequests, and then it would be convenient of someone could install the flatpack version next to the regular version and have them completely isolated.

README.md
```diff
@@ -178,7 +227,31 @@ run:
...
+
+##### Migrating data to flatpak
+
+If you would like to retain your data from a non-flatpak installation,
+you can do so by running:
+
+```bash
+gio copy -b \
+    ~/.local/share/hamster/hamster.db \
+    ~/.var/app/org.gnome.Hamster/data/hamster/
```
matthijskooijman, line 249: Why use `gio copy` here? Looks like these are two regular paths that can just use normal `cp`?

README.md
mwilck: > That does make me wonder: What does flatpack do when the dbus name is already registered (i.e. when hamster is normally installed as well)? From a quick test it seems to (silently) not claim the name, that is, when I start the flatpack, the flatpack GUI connects to the non-flatpack dbus service, and when I kill the non-flatpack services, it still spawns the non-flatpack ones, rather than the flatpack ones.

I don't think it's possible to hijack a well-known dbus address. I guess it shouldn't be. At best, hamster itself should log a meaningful message when it fails to acquire the address. But that's independent of packaging. It applies also to a situation with installed hamster and locally (developer) instance, for example.

It's a general problem that I have with all applications that can be installed either "regularly" or as flatpak/snap/appimage (you name it): It's highly intransparent to the user which package she is activating. DBus makes it even more intransparent, of course.

matthijskooijman: > I don't think it's possible to hijack a well-known dbus address. I guess it shouldn't be. At best, hamster itself should log a meaningful message when it fails to acquire the address. But that's independent of packaging. It applies also to a situation with installed hamster and local (developer) instance, for example.

Well, there's two parts to this: One is the claiming of a dbus name at runtime, which is something hamster should report. The other, which I was thinking of, is installing a file in `/usr/share/somewhere` that declares that a dbus interface exists and what binary to autostart if the name is requested but not on the bus yet (at least that's how I gather it works). This is something that `flatpack install` should handle, but maybe flatpack handles this differently? Also, it [seems that the dbus name should match the flatpak name](https://github.com/flatpak/flatpak/issues/1902#issuecomment-405049241), not sure if our `WindowServer` postfix is then supported (the primary service does match).

salim-b:
Just a follow-up question.

The README now says:

> If you downloaded the file with the Hamster bundle (ending in `.flatpak`), you can directly install it with:
> ```
> flatpak install --reinstall Hamster.flatpak
> ```


But it **doesn't** say _where_ you can download the Flatpak. I guess you are referring to the corresponding [GitHub Actions artifact](https://github.com/projecthamster/hamster/suites/1020590782/artifacts/13398291)[1]?

Any plans on distributing the Flatpak on [Flathub](https://flathub.org/)?

[1]: I've succesfully tested this in conjunction with the GNOME-3.36 version of [hamster-shell-extension](https://github.com/projecthamster/hamster-shell-extension/issues/327) on Ubuntu 20.04, works like a charm. :+1:

tchernobog:
> But it doesn't say where you can download the Flatpak. I guess you are referring to the corresponding GitHub Actions artifact[1]?

You can get it from there for now. I would assume  the GNOME Hamster maintainers could promote the built .flatpak file to a release artifact for next release.

> Any plans on distributing the Flatpak via Flathub?

That would require to get the manifest merged to a different repo, see: https://github.com/flathub/flathub/wiki/App-Submission
I could do so if there are enough requests, but first I would like to get the green light from the app maintainers, as they should own the new repository.

Issue #615, opened 2020-05-29
Activity search behaves strangely with dash in search term


mwilck:
Searching for "suse-module-tools" in my activity list, I get a bunch of false matches:

![Screenshot from 2020-05-29 16-47-45](https://user-images.githubusercontent.com/15961354/83273130-6cdc5200-a1cc-11ea-8693-67a32789879e.png)

It seems to match with the `SUSE` category name. Still I find this rather not intuitive with the search term "`suse-module-tools`". I'm currently running hamster e6b90fc, which doesn't differ from 3.0.2 in any significant way for this problem, AFAICS.

 (Those black bars look strange as well, but that's just a nitpick).

GeraldJansen:
I have seen the same behaviour searching for 'x/y', however 'x.y' does pick up 'x/y'. You could try 'suse.module.tools'. It may be searching for regex's, but I'm not sure about that.

esteban-aliverti:
I think it would be nice to define what the expected behavior is so somebody could start working on a solution.

GeraldJansen:
AFAIKS the search is simply performed using an Sqlite3 LIKE clause ([documentation](https://sqlite.org/lang_expr.html#like)), so nothing to do with regex. Sorry for the noise.

GeraldJansen:
It seems to me, from casual testing, that "punctuation" characters like '/', '.' and '-' are simply ignored.

PR #619, opened 2020-06-13
Remove dark bars for empty days in overview.

GeraldJansen:
The overview currently has dark bars for days with no recorded facts.  These are considered unsightly by some (eg. #615, and me too). The behaviour is inconsistent because empty days at the start and end of the date range are not highlighted whereas intermediate empty days are.

This PR completely skips over empty days in the selected date range.

### DIFF
--- a/src/hamster/widgets/facttree.py
+++ b/src/hamster/widgets/facttree.py
@@ -514,7 +514,8 @@ class FactTree(graphics.Scene, gtk.Scrollable):
         days = []
         for i in range((end - start).days + 1):
             current_date = start + dt.timedelta(days=i)
-            days.append((current_date, by_date[current_date]))
+            if current_date in by_date:
+                days.append((current_date, by_date[current_date]))
 
         self.days = days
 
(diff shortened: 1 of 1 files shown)


GeraldJansen:
@matthijskooijman, please help me with the failed "code linting" test.  I made a separate commit in to fix the flake8 issues in #620, and I guess they have also been fixed in another merged PR. Do I have to rebase?

matthijskooijman:
Yup, rebasing would be a good approach here to get the changes from the merged flake8 fixes, without needlessly complicating history like with a regular merge. I suspect you might run into a small conflict (because this PR touches the same file and also makes a whitespace-only change I believe, so maybe that same change was made in master too), but that should be fairlly easy to fix. After the rebase, you'll need to do a force push to replace the existing commit.

Let me know if you need more specific guidance.

GeraldJansen:
Okay, rebase done.

GeraldJansen:
Closing out of impatience. Bye.

Issue #568, opened 2020-02-19
3.0-rc1: hamster ignores screen lock events


mwilck:
Hamster 2.2 would catch screen lock events and stop the current activity. This was a very important feature - without it, users need to remember manually stopping tracking when  take a break or otherwise leave the screen. Even if the activation of the screen lock doesn't necessarily mean that activities have stopped, it's very helpful to have the time stamp of the onset of idleness recorded this way, to be able to manually correct activities later.

This regression has been caused by cc96625 ("remove idle.py"). The commit contains no explanation why this important functionality has been removed. My guess is it was because the `idle.py` code was accessing the deprecated gconf functionality. However, 

  1. This was only used to calculate backwards from the actual "lock" event towards the time when the desktop last recorded any activity,
 2. The gsettings replacement would be `lock-delay` in `/org/gnome/desktop/screensaver`.

mwilck:
Well, there's more:

 * 5f2f061 remove stop_on_shutdown preference
 * cc96625 remove idle.py
 * 08eaf77 remove desktop.py
 * 487a46f remove notify_interval preference
 * 7e5b687 remove notify_on_idle preference
 * 891d33b remove enable_timeout preference
 * 0605cb3 disable desktop integration

@ederag, why have you ditched all this functionality? Where's the replacement?

GeraldJansen:
From #494:
> Remove unused external.py or currently nonfunctional stuff (desktop/idle)
> that should be moved to a companion project, as discussed in #493.

mwilck:
#493 talks about the "external" functionality only. #494 provides no insight to me, other than that it claims that "idle" wasn't functional, which is not true, at least not for hamster 2.2.

mwilck:
To put it differently: if "idle" was broken before #494, it must have been broken by some other changes made after 2.2.2.

GeraldJansen:
In addition to the considerable insight provided by #493, in my opinion, please keep in mind that Hamster is also used (and developed and tested) in non-Gnome environments. Furthermore, the project is actually pretty short on active testers and contributors from among the Gnome users. Just to voice a contrasting opinion, I for one think that @ederag made the right decision to "ditch" all that external stuff, given the lack of resources. On a positive note, it looks like @sh-zam plans to build an external service to restore some of that functionality.

mwilck:
> In addition to the considerable insight provided by #493,

Sorry, I'm unable to find any insight about idle detection in #493.

[...]

PR #631, opened 2020-09-20
Restore shutdown

sgranjoux:
Restore the ability to stop the current task on log out.

Add a new class to listen for GNOME session events on DBus.
Restore shutdown check box,

### DIFF
--- a/data/preferences.ui
+++ b/data/preferences.ui
@@ -41,6 +41,22 @@
                     <property name="orientation">vertical</property>
                     <property name="spacing">8</property>
                     <child>
+                      <object class="GtkCheckButton" id="shutdown_track">
+                        <property name="label" translatable="yes">Stop tracking on shutdown</property>
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="receives_default">False</property>
+                        <property name="use_underline">True</property>
+                        <property name="xalign">0.5</property>
+                        <property name="draw_indicator">True</property>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">True</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child>
                       <object class="GtkBox" id="day-start box">
                         <property name="visible">True</property>
                         <property name="can_focus">False</property>
@@ -54,7 +70,7 @@
                             <property name="expand">False</property>
                             <property name="fill">True</property>
                             <property name="padding">4</property>
-                            <property name="position">0</property>
+                            <property name="position">1</property>
                           </packing>
                         </child>
                         <child>
--- a/src/hamster-service.py
+++ b/src/hamster-service.py
@@ -13,6 +13,7 @@ from hamster.lib import i18n
 i18n.setup_i18n()  # noqa: E402
 
 from hamster.storage import db
+from hamster.session import DbusSessionListener
 from hamster.lib import datetime as dt
 from hamster.lib import default_logger
 from hamster.lib.dbus import (
@@ -26,6 +27,7 @@ from hamster.lib.dbus import (
     to_dbus_fact_json
 )
 from hamster.lib.fact import Fact, FactError
+from hamster.lib.configuration import conf
 
 logger = default_logger(__file__)
 
@@ -38,7 +40,7 @@ if "org.gnome.Hamster" in dbus.SessionBus().list_names():
     quit()
 
 
-class Storage(db.Storage, dbus.service.Object):
+class Storage(db.Storage, dbus.service.Object, DbusSessionListener):
     __dbus_object_path__ = "/org/gnome/Hamster"
 
     def __init__(self, loop):
@@ -50,6 +52,7 @@ class Storage(db.Storage, dbus.service.Object):
         db.Storage.__init__(self, unsorted_localized="")
 
         self.mainloop = loop
+        DbusSessionListener.__init__(self)
 
         self.__file = gio.File.new_for_path(__file__)
         self.__monitor = self.__file.monitor_file(gio.FileMonitorFlags.WATCH_MOUNTS | \
@@ -78,6 +81,12 @@ class Storage(db.Storage, dbus.service.Object):
                 for activity in activities:
                     self.add_activity(activity, cat_id)
 
+
+    # stop current task on log out if requested
+    def end_session(self):
+        """Stop tracking on logout."""
+        if conf.get("stop-on-shutdown"):
+            self.stop_tracking(None)
 
     # stop service when we have been updated (will be brought back in next call)
     # anyway. should make updating simpler
--- a/src/hamster/session.py
+++ b/src/hamster/session.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (C) 2020 Sébastien Granjoux <seb.sfo@free.fr>
+
+# This file is part of Project Hamster.
+
+# Project Hamster is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# Project Hamster is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with Project Hamster.  If not, see <http://www.gnu.org/licenses/>.
+
+import dbus
+
+class DbusSessionListener(object):
+    """Listen for GNOME manager end session event."""
+
+
+    def end_session(self):
+        """Override this method to do something at the end of a session.
+        It must not attempt to interact with the user and will be given
+        a maximum of ten seconds to perform the actions."""
+        pass
+
+
+    def __init__(self):
+        """Connect to the GNOME manager session signals"""
+
+        # Get SessionManager interface
+        session_bus = dbus.SessionBus()
+        session_manager = session_bus.get_object("org.gnome.SessionManager",
+            "/org/gnome/SessionManager")
+        self.__session_manager_iface = dbus.Interface(session_manager,
+            dbus_interface="org.gnome.SessionManager")
+        self.__client_id = self.__session_manager_iface.RegisterClient("", "")
+
+        # Get SessionManager.ClientPrivate interface
+        session_client = session_bus.get_object("org.gnome.SessionManager",
+            self.__client_id)
+        self.__session_client_private_iface = dbus.Interface(session_client,
+            dbus_interface="org.gnome.SessionManager.ClientPrivate")
+
+        # Connect to the needed signals
+        session_bus.add_signal_receiver(self.__query_end_session_handler,
+            signal_name = "QueryEndSession",
+            dbus_interface = "org.gnome.SessionManager.ClientPrivate",
+            bus_name = "org.gnome.SessionManager")
+
+        session_bus.add_signal_receiver(self.__end_session_handler,
+            signal_name = "EndSession",
+            dbus_interface = "org.gnome.SessionManager.ClientPrivate",
+            bus_name = "org.gnome.SessionManager")
+
+        session_bus.add_signal_receiver(self.__stop_handler,
+            signal_name = "Stop",
+            dbus_interface = "org.gnome.SessionManager.ClientPrivate",
+            bus_name = "org.gnome.SessionManager")
+
+
+    def __query_end_session_handler(self, flags):
+        """Inform that the session is about to end. It must reply with
+        EndSessionResponse within one second telling if it is ok to proceed
+        or not and why. If flags is true, the session will end anyway."""
+        self.__session_client_private_iface.EndSessionResponse(True, "")
+
+    def __end_session_handler(self, flags):
+        """Inform that the session is about to end. It must reply with
+        EndSessionResponse within ten seconds."""
+        self.end_session()
+        self.__session_client_private_iface.EndSessionResponse(True, "")
+
+    def __stop_handler(self):
+        """Remove from the session."""
+        self.__session_manager_iface.UnregisterClient(self.__client_id)
--- a/data/org.gnome.hamster.gschema.xml
+++ b/data/org.gnome.hamster.gschema.xml
@@ -7,6 +7,14 @@
                 The folder the last report was saved to
             </description>
         </key>
+        
+        <key type="b" name="stop-on-shutdown">
+            <default>false</default>
+            <summary>Stop tracking on shutdown</summary>
+            <description>
+                Stop tracking current activity on shutdown
+            </description>
+        </key>
 
         <key type="u" name="day-start-minutes">
             <default>330</default>
--- a/src/hamster/preferences.py
+++ b/src/hamster/preferences.py
@@ -169,6 +169,8 @@ class PreferencesEditor(Controller):
         self.window.show_all()
 
     def load_config(self, *args):
+        conf.bind("stop-on-shutdown", self.get_widget("shutdown_track"), "active")
+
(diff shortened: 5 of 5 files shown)


mwilck:
I haven't looked at details yet, but in general this is highly welcome. I wonder if you've also looked into stopping tracking when the lock screen is activated? Have you seen #568 and the related discussion?

sgranjoux:
I'm using hamster since a long time and I'm missing this 'stop tracking on shutdown` feature. But I haven't much time to work on this so it takes a quite long time to annoy me enough to take more active actions.

I have take a quick look at the 'stop tracking when idle' feature when writing this. It looks like it could be restored too but I haven't planned to do it, at least in a short future.

I have just read #568 and a few related discussions. Well, at least there are some life, we will see how it goes.

matthijskooijman:
I've had a look at this PR. I like the idea of restoring this behavior.

What confuses me, though, is that one commit talks about shutdown and the other about logout, but in practice I think the feature applies to both (i.e. any clean termination of the gnome session), right?

Also, it seems this PR is very Gnome-specific? IMHO it's acceptable to only support a subset of environments (i.e. only Gnome right now maybe), but what happens with the current code on e.g. KDE? Is there (I believe not?) any way for the user to know that the feature is not supported, or is the checkbox in the settings just shown but ineffective? That would not be acceptable IMHO. Also, would the lack of a GnomeSession DBUS service be handled gracefully?

I wonder if we cannot detect logout in a desktop-environment-agnostic way? If the hamster daemon is running inside the user session, it will be terminated as part of logout, so couldn't that just be a moment to stop tracking? Or is that hard to distinguish from manually stopping hamster (but does that ever happen)?

Alternatively, maybe systemd-logind can help here? That might work across different desktop environments and maybe even allow detecting any secondary sessions for the same user, if we think that's relevant. Though logind might not be enabled on all systems, of course.

sgranjoux:
Indeed, it stop the tracking on logout. I have kept shutdown because I think it was the name used before this feature has been removed.

I have not checked what's happen on KDE, I think it will just do nothing.

I think I have checked how this feature has been implemented and just replace it with something similar. I haven't found a general way to achieve that. I think it's fine if we stop the tracking when the user stop the daemon but I don't know if stopping the tracking won't need another user process which can be already stopped.

Then using systemd-logind or something else for KDE will probably not work on all systems, so in any case we need to check what's happen when the DBUS service is missing and probably disable this option.

GeraldJansen:
This definitely breaks Hamster on non-Gnome systems. I've tried on XFCE and get this vomit:

```
[hamster]$ hamster
hamster-service up
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 177, in activate_name_owner
    return self.get_name_owner(bus_name)
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 361, in get_name_owner
    return self.call_blocking(BUS_DAEMON_NAME, BUS_DAEMON_PATH,
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner: Could not get owner of name 'org.gnome.SessionManager': no such name

During handling of the above exception, another exception occurred:

  File "/d/gj/hamster/src/hamster-service.py", line 467, in <module>
    storage = Storage(loop)
  File "/d/gj/hamster/src/hamster-service.py", line 55, in __init__
    DbusSessionListener.__init__(self)
  File "/d/gj/hamster/src/hamster/session.py", line 38, in __init__
    session_manager = session_bus.get_object("org.gnome.SessionManager",
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 241, in get_object
    return self.ProxyObjectClass(self, bus_name, object_path,
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 250, in __init__
    self._named_service = conn.activate_name_owner(bus_name)
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 182, in activate_name_owner
    self.start_service_by_name(bus_name)
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 277, in start_service_by_name
    return (True, self.call_blocking(BUS_DAEMON_NAME, BUS_DAEMON_PATH,
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.SessionManager was not provided by any .service files
ERROR:dbus.proxies:Introspect error on :1.522:/org/gnome/Hamster: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
  File "/d/gj/hamster/src/hamster-cli.py", line 149, in on_activate_window
  File "/d/gj/hamster/src/hamster-cli.py", line 184, in _open_window
    self.overview_controller = Overview()
  File "/d/gj/hamster/src/hamster/overview.py", line 477, in __init__
    self.find_facts()
  File "/d/gj/hamster/src/hamster/overview.py", line 531, in find_facts
    self.facts = self.storage.get_facts(start, end, search_terms=search)
  File "/d/gj/hamster/src/hamster/client.py", line 156, in get_facts
    for fact in self.conn.GetFactsJSON(dbus_range, search_terms)]
  File "/d/gj/hamster/src/hamster/client.py", line 99, in conn
    server_version = self._connection.Version()
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 72, in __call__
    return self._proxy_method(*args, **keywords)
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name :1.522 was not provided by any .service files
ERROR:dbus.proxies:Introspect error on :1.524:/org/gnome/Hamster: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
  File "/d/gj/hamster/src/hamster-cli.py", line 149, in on_activate_window
  File "/d/gj/hamster/src/hamster-cli.py", line 184, in _open_window
    self.overview_controller = Overview()
  File "/d/gj/hamster/src/hamster/overview.py", line 477, in __init__
    self.find_facts()
  File "/d/gj/hamster/src/hamster/overview.py", line 531, in find_facts
    self.facts = self.storage.get_facts(start, end, search_terms=search)
  File "/d/gj/hamster/src/hamster/client.py", line 156, in get_facts
    for fact in self.conn.GetFactsJSON(dbus_range, search_terms)]
  File "/d/gj/hamster/src/hamster/client.py", line 99, in conn
    server_version = self._connection.Version()
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 72, in __call__
    return self._proxy_method(*args, **keywords)
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name :1.524 was not provided by any .service files
```

sgranjoux:
Thank you for checking that, I have not much time but I will fix that.

sgranjoux:
I have caught the exception and added a warning if the GNOME session is not available. The I have added a tool tip on the setting to warn the user that this function is not available outside GNOME. It was pretty straightforward.

I have put all this in a third commit. Do you prefer that I merge those changes in the previous two commits?

I have kept the name shutdown because it was the old name. Do you prefer that I call that "Stop tracking on logout"?

Then for other environments. I'm afraid that stopping the tracking when the hamster daemon is terminated by the log out could lead to nasty bug depending on the order in which other processes are killed. Do you think it can be an issue?

Else, we can probably get an end session signal on KDE and perhaps XFCE or by using systemd. Do you have a preference?

src/hamster/session.py
```diff
@@ -35,34 +37,36 @@ def __init__(self):
...
+                signal_name = "EndSession",
+                dbus_interface = "org.gnome.SessionManager.ClientPrivate",
+                bus_name = "org.gnome.SessionManager")
+
+            session_bus.add_signal_receiver(self.__stop_handler,
+                signal_name = "Stop",
+                dbus_interface = "org.gnome.SessionManager.ClientPrivate",
+                bus_name = "org.gnome.SessionManager")
+        except dbus.exceptions.DBusException:
+            logger.warning("Unable to connect to GNOME session manager, stop trcking on logout won't work")
```
GeraldJansen, line 69: Please either make this logger.info() or just replace this line with `pass` so command line users don't see this message on every single command (because warning is the default logging level). Also trcking -> tracking.

data/preferences.ui
```diff
@@ -49,6 +49,8 @@
                         <property name="use_underline">True</property>
                         <property name="xalign">0.5</property>
                         <property name="draw_indicator">True</property>
+                        <property name="has_tooltip">True</property>
+                        <property name="tooltip_text" translatable="yes">Work only on GNOME</property>
```
GeraldJansen, line 53: "Only works on GNOME"

GeraldJansen:
Tried again on XFCE and no longer seems to cause problems. I didn't notice any problem when killing the hamster daemon from the command line or when logging out (i.e. an ongoing activity remains open) . For me it is not a problem to introduce functionality available only on GNOME.

data/preferences.ui
GeraldJansen: Actually I think it would be better if the main description was changed to "Stop tracking on shutdown (GNOME only)".

src/hamster/session.py
sgranjoux: Ok, I have changed it but I think this message will appear only when your start the hamster daemon so probably not on each command.

data/preferences.ui
sgranjoux: The tooltip appears quite easily here so I think it was enough but I have changed the label.

src/hamster/session.py
GeraldJansen: Ah right, I forgot that I was using a cover script that restart the daemon before each command. Thanks anyway.

GeraldJansen:
> I wonder if we cannot detect logout in a desktop-environment-agnostic way?

Maybe this should be a new issue / feature request. Perhaps one could listen for [org.freedesktop.login1](https://www.freedesktop.org/software/systemd/man/org.freedesktop.login1.html) PrepareForShutdown and PrepareForSleep signals?

src/hamster/session.py
```diff
@@ -0,0 +1,85 @@
...
+                dbus_interface = "org.gnome.SessionManager.ClientPrivate",
+                bus_name = "org.gnome.SessionManager")
+        except dbus.exceptions.DBusException:
+            logger.info("Unable to connect to GNOME session manager, stop tracking on logout won't work")
+
+    def __query_end_session_handler(self, flags):
+        """Inform that the session is about to end. It must reply with
+        EndSessionResponse within one second telling if it is ok to proceed
+        or not and why. If flags is true, the session will end anyway."""
+        self.__session_client_private_iface.EndSessionResponse(True, "")
```
matthijskooijman, line 75: Why is this handler needed? If we're not doing anything other than saying it is ok to end the session, isn't this just the default? Or is it so that *if* you register something (I'm not sure about right terminology here) against SessionManager, that you *need* to also handle this QueryEndSession event?

src/hamster/session.py
```diff
@@ -0,0 +1,85 @@
...
+
+class DbusSessionListener(object):
+    """Listen for GNOME manager end session event."""
+
+
+    def end_session(self):
+        """Override this method to do something at the end of a session.
+        It must not attempt to interact with the user and will be given
+        a maximum of ten seconds to perform the actions."""
+        pass
```
matthijskooijman, line 32: Maybe this should raise `NotImplementedError`?

src/hamster-service.py
```diff
@@ -38,7 +40,7 @@
     quit()
 
 
-class Storage(db.Storage, dbus.service.Object):
+class Storage(db.Storage, dbus.service.Object, DbusSessionListener):
```
matthijskooijman, line 43: I wonder if adding a mixin class here is the cleanest way to implement this. Also, it doesn't seem a responsibility of a class called Storage to implement a "stop-tracking-on-shutdown", but maybe that class already does more than its name implies (haven't looked too closely now).

matthijskooijman (commented):
I had a look over the code. It looks ok overall, left some inline remarks (mostly questions). I haven't tested the code.

> Maybe this should be a new issue / feature request. Perhaps one could listen for org.freedesktop.login1 PrepareForShutdown and PrepareForSleep signals?

Agreed, I do no think this must be solved before merging this, provided that this limitation if sufficiently clear for the user.

On that track, I like mentioning it is GNOME-only in the label more than the tooltip, since people might not be reading the tooltip and then be surprised it doesn't work.

> I have kept the name shutdown because it was the old name. Do you prefer that I call that "Stop tracking on logout"?

I personally would find "shutdown" confusing, since that implies a system shutdown to me.

Maybe "Stop tracking when GNOME session ends" could be an alternative, which neatly also includes the GNOME limitation?


> Then for other environments. I'm afraid that stopping the tracking when the hamster daemon is terminated by the log out could lead to nasty bug depending on the order in which other processes are killed. Do you think it can be an issue?

What issue are you afraid of exactly? I can imagine that whenever the hamster process is terminated, it could just stop tracking. This probably needs a SIGTERM handler or something like that, though maybe Python has some "on process end" code (or maybe the mainloop already neatly stops running on a SIGTERM and you could just add some code after the mainloop?).

I still think it would be good to consider this alternative, since if it would work, would not only work on non-GNOME systems, but also on GNOME systems, so I expect that the code for this would be simpler and maybe less fragile than listening on the gnome session dbus.

One thing that might prevent this from working properly, though, is that it relies on the hamster daemon being stopped when the login session ends, which might not actually be the case. I.e. there is systemd-logind, which I think supports processes that are tied to a user, not a particular login session (so they are killed only when the last login session of that user is logged out). I'm not sure how that works with dbus-autostart, though (which I think is how hamster is being started now in most cases? Or is it a .desktop file in an autostart directory?).

src/hamster/session.py
sgranjoux: I have not checked, but according to the comment I have understood that when you register against the SessionManager you have to handle this message. 
If it's not done, I think it can delay the log out by 1 seconds, eventually the session manager could remove the corresponding program and we will not get the end session message.

src/hamster/session.py
sgranjoux: I think pass is fine, there is no need to do something.
Eventually, I could extend this object to provide a hook in query_end_session_handler. I will allow to use it to delay the end of a session and it could be that nothing is needed when the session end. Anyway, it's not done currently, so I could raise `NotImplementedError` if you think it's better.

src/hamster-service.py
sgranjoux: I agree with that. It's probably the simplest way but not the cleanest one.

I think I can create a separate object having a reference on the storage object to call StopTracking. To make it even more separated, this object can send a stop tracking message on the dbus, it could be even a separate process.

sgranjoux:
> Maybe "Stop tracking when GNOME session ends" could be an alternative, which neatly also includes the GNOME limitation?
That's fine for me. 

Should I change the name of the gconf key from `stop-on-shutdown` to `stop-on-logout`?


> What issue are you afraid of exactly?

When you log out, all user programs are killed, I think it includes supporting processes used to exchange dbus signal or read gconf key. I'm afraid that those functions are used by hamster and don't work at that time but I haven't checked anything. 

> One thing that might prevent this from working properly, though, is that it relies on the hamster daemon being stopped when the login session ends, which might not actually be the case. I.e. there is systemd-logind, which I think supports processes that are tied to a user, not a particular login session

I don't think it's a problem is the tracking is not stop in such case.

Issue #639, opened 2020-11-15
Cannot change date range in overview


taniwallach:
I installed on Debian buster using the current Git repository.

When hamster is open, I cannot "pull down" the date from the top menu to set a range of dates.

Navigation to different single days using the "<" and ">" icons does work.

I tried the snap also, and the had the same problem with it.

GeraldJansen:
Hmm, strange. What do you mean by "pull down"? Just clicking on the date should open the popup. (I just tried current git on my Debian testing and it seems to work fine.)

taniwallach:
There is a slight flicker, but nothing opens:
[Hamster-Debian10-Click-Date-no-reaction.zip](https://github.com/projecthamster/hamster/files/5544829/Hamster-Debian10-Click-Date-no-reaction.zip)

Maybe a missing depencency?

Maybe the "window service" is not starttng, as there was some mention of a need to sometimes kill 2 services, and I see only one:
```
tani@lxtani:/var/log$ ps -aux | grep hamster
tani     14975  0.0  0.2 301036 39012 ?        Sl   Nov15   0:08 python3 /usr/libexec/hamster/hamster-service
tani     15017  0.0  0.3 502800 60864 ?        Sl   Nov15   0:12 python3 /usr/bin/hamster --gapplication-service
tani     19992  0.0  0.0   7140   892 pts/1    S+   09:55   0:00 grep hamster
```

GeraldJansen:
I also no longer see hamster-window-service running with recent versions. However, I do not get the --gapplication-service command line argument. You could try killing hamster and hamster-service, then from the command line try 
`/usr/bin/hamster version`, which should give v3.0.2-8-gceedcd87, and then try `/usr/bin/hamster &`.  

Sorry for the noise. Hopefully someone more knowledgeable about recent developments (eg. snap install) can help.

taniwallach:
I think I have a fix in https://github.com/projecthamster/hamster/pull/640. It works for me. The "popup" opens/closes when the button is clicked. Click on the Today/Week/Month line, or on Apply makes the setting and closes the "popup" window.

The changes can be applied for testing on a Debian system by making the changes to `/usr/lib/python3/dist-packages/hamster/widgets/dates.py` and the running
```
pkill -f hamster-service ; /usr/bin/hamster
```

taniwallach:
[...]

PR #640, opened 2020-11-16
Fix RangePick - was not displaying on Debian 10

taniwallach:
Fixes https://github.com/projecthamster/hamster/issues/639 on Debian 10, at least for me.


taniwallach:
Better fixes are in the patches in https://github.com/projecthamster/hamster/issues/639 and once something settles down,  I expect that this issue will be fixed in a better way by a different PR.

PR #647, opened 2020-11-19
Simplify Rangepick widget to fix issues on Debian Buster

### LINKED ISSUE #645: Issue of popup position of RangePick on Debian 10 with wayland
Originally discussed in: https://github.com/projecthamster/hamster/issues/639.

On Debian 10 with wayland: the popup always opens in the top left of the screen, and not placed over the hamster window (offset down and right a bit).

---

Le mercredi 18 novembre 2020, Nathan Wallach a écrit :
> It is possible that the data below hints that these issues are related to one of:
>   - `xorg` (was in use on the machine where there was an issue) vs `wayland` (in use on the machine where things worked) **seems more likely**

I confirm that a combination of wayland + not very recent GTK version
creates this issue where some popups windows are misplaced. I got the same
problem in another application (tryton).

Cheers,
-- 
Raphaël Hertzog

_Originally posted by @rhertzog in https://github.com/projecthamster/hamster/issues/639#issuecomment-729788375_

---
Some issues of popup positioning with wayland were fixed by https://github.com/projecthamster/hamster/pull/527. Perhaps the same fix is needed here.

_Originally posted by @GeraldJansen in https://github.com/projecthamster/hamster/issues/639#issuecomment-729796719_

matthijskooijman:
This simplifies the Rangepick widget by using `gtk.Popover` and `gtk.MenuButton`, conveniently handle a lot of the boilerplate of hiding and showing the popup automatically. This also allows removing some problematic code, fixing #639 and because no separate window is created anymore, this indirectly also fixes #645.

This slightly changes the looks of the popup, previously it was:

image

Now it is:

image

Note the "arrow", border and slightly different positioning of the popup.

One downside of using a Popover is that it is a widget, rather than an window, so at least on X11, it cannot extend outside of the containing window, but this is probably not a problem in practice (this limitation does not seem to exist on Wayland, see https://gitlab.gnome.org/GNOME/gtk/-/issues/543).

This PR supersedes #643 and as far as I'm concerned, is how we should be fixing these issues.

@taniwallach, would you be so kind to give this PR a final test on your two machines?

@others, anyone want to give this PR a technical review, so we can get it merged?

### DIFF
--- a/data/date_range.ui
+++ b/data/date_range.ui
@@ -2,14 +2,8 @@
 <!-- Generated with glade 3.16.1 -->
 <interface>
   <requires lib="gtk+" version="3.2"/>
-  <object class="GtkWindow" id="range_popup">
+  <object class="GtkPopover" id="range_popup">
     <property name="can_focus">False</property>
-    <property name="resizable">False</property>
-    <property name="window_position">mouse</property>
-    <property name="destroy_with_parent">True</property>
-    <property name="skip_taskbar_hint">True</property>
-    <property name="skip_pager_hint">True</property>
-    <property name="decorated">False</property>
     <child>
       <object class="GtkFrame" id="frame1">
         <property name="visible">True</property>
--- a/src/hamster/widgets/dates.py
+++ b/src/hamster/widgets/dates.py
@@ -93,6 +93,7 @@ class RangePick(gtk.ToggleButton):
         self._ui = load_ui_file("date_range.ui")
 
         self.popup = self.get_widget("range_popup")
+        self.popup.set_relative_to(self)
 
         self.today = today
 
@@ -106,12 +107,11 @@ class RangePick(gtk.ToggleButton):
         self.start_date, self.end_date = None, None
         self.current_range = None
 
-        self.popup.connect("focus-out-event", self.on_focus_out)
+        self.popup.connect("closed", self.on_closed)
         self.connect("toggled", self.on_toggle)
 
         self._ui.connect_signals(self)
         self.connect("destroy", self.on_destroy)
-        self._hiding = False
 
     def on_destroy(self, window):
         self.popup.destroy()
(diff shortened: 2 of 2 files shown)


rhertzog (approved):
I tested it on Debian unstable and it works fine for me, also fixing the bad placement of the window. Nice simplification!

taniwallach:
I won't have access to the specific second machine used before untill I am back in the office on Sunday, but set up a different second machine to test on.

The code in this PR was tested on the laptop (Debian 10, xorg) which had the issue with no RangePick popup opening with the current master branch, and on a very clean Debian 10 using Wayland (checked and hits the popup location issue with master).

All functionality of the RangePick popup works as expected on both of these machines.

taniwallach (approved):
Works for me and fixes the issues experienced before on the 2 Debian 10 machines (xorg / wayland).

I'm not experience enough with python or GTK to really review the code, but the GtkPopover implementation code is certainly simple and cleaner that the prior GtkWindow code for this functionality.

GeraldJansen (commented):
Very nice simplifications.

matthijskooijman:
Given @rhertzog and @GeraldJansen have left positive comments (even though not proper "approve" reviews), I think this can be merged.

I've rebased this on top of master and will merge next.

rhertzog:
@matthijskooijman I can't do proper "approve" reviews. I'm not a member of the github project and I don't have the required rights. I proposed my (occasional) help to join the team but never got any feedback.

matthijskooijman:
>  I can't do proper "approve" reviews.

Ah, I thought that anyone could do that. Maybe that's a setting that @elbenfreund could set.

> I proposed my (occasional) help to join the team but never got any feedback.

We don't really have any proper process for that yet, but I've asked other maintainers for their thoughts just this morning, so we'll get back to you soon :-)

Issue #607, opened 2020-05-13
Automate NEWS generation


matthijskooijman:
The NEWS has a high-level changelog, which is currently manually written at release time (or sometimes in chunks between releases).

I would like to investigate a scheme were changelog entries are written along with the code changes (i.e. as part of the pull request). An obvious way would be to just modify the NEWS file in a commit inside the PR, but that is going to cause a lot of conflicts, and makes it harder to include the PR number in the changelog).

An obvious way would be to just use the git log to generate the changelog, but IMHO that would produce a too finegrained changelog. e.g. for NEWS it useful to know a bug was fixed, but there is no need to know that three refactoring commits were needed before that. Also, if you want the git log, you can just use git or github to see it.

So ideally, you would explicitly *add* a message to each PR description (or selected commit messages), which are then automatically generated into a NEWS file at releasetime with some script (probably with some chance to do manual post-processing if needed).

It seems there are quite some tools for this. Reading through [this article](https://depfu.com/blog/changelogs-to-write-or-to-generate), Changelox looked promising, but their website is giving an SSL failure right now. The github-changelog-generator project [also has a list of competitors](https://github.com/github-changelog-generator/github-changelog-generator/wiki/Alternatives) that might prove useful.

For the upcoming 3.0.2 release, the changelog is simple enough to just do it manually as before, but I'd like to see if we can set up something like this for the next release cycle.

matthijskooijman:
I just remembered I recently saw some other repository where they collected changelog entries in separate files in the repository, to be removed and merged into a NEWS file on release. Putting things in separate files, rather than in a single file helps to prevent git conflicts, so that's also a thing to consider (but it does suffer from the same problems with not being to include PR numbers easily).

This also made me realize one more requirement: It should be possible to *update* changelog entries between the moment they are created and the release. You might find a typo, or want to add some additional information later. You can of course do this at release time, but you are likely to forget that unless we have a mechanism to remember these changes. It would be better if such a change could be made immediately, e.g. by editing the PR description or editing a file in the repo. This is probably not possible when putting changelog entries in commit messages.

matthijskooijman:
Then, it might be useful if there was some automatic categorization of changelog entries (e.g. to split off user-visible changes, internal improvements, build system changes, etc.). Also, this could allow marking (breaking) API changes (e.g. in dbus, python imports or commandline) separately and having those in a separate list in the changelog (this might mean that a single change needs to appear in two lists as well, so that should also be possible).

[...]

PR #648, opened 2020-11-19
Reset position in FactTree only when needed

### LINKED ISSUE #594: Activity list jumps to top after some short time

rhertzog:
Right now FactTree.set_facts() is always resetting the position of the
view, bringing the user back to the top but this function is called
regularly to update with the latest data even when the user has not
changed anything in the selection widget... which is really annoying
when you are reviewing the list of facts.

Let's distinguish between both cases with a "reset=True" attribute that
we use when we generate an entirely new set of facts to be displayed.

This makes #623 obsolete. Thanks to @mwilck for the initial analysis.

Fixes #594

### DIFF
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -523,19 +523,19 @@ class Overview(Controller):
         if event.keyval == gdk.KEY_Escape:
             self.close_window()
 
-    def find_facts(self):
+    def find_facts(self, scroll_to_top=False):
         start, end = self.header_bar.range_pick.get_range()
         search_active = self.header_bar.search_button.get_active()
         search = "" if not search_active else self.filter_entry.get_text()
         search = "%s*" % search if search else "" # search anywhere
         self.facts = self.storage.get_facts(start, end, search_terms=search)
-        self.fact_tree.set_facts(self.facts)
+        self.fact_tree.set_facts(self.facts, scroll_to_top=scroll_to_top)
         self.totals.set_facts(self.facts)
         self.header_bar.stop_button.set_sensitive(
             self.facts and not self.facts[-1].end_time)
 
     def on_range_selected(self, button, range_type, start, end):
-        self.find_facts()
+        self.find_facts(scroll_to_top=True)
 
     def on_search_changed(self, entry):
         if entry.get_text():
@@ -544,7 +544,7 @@ class Overview(Controller):
         else:
             self.filter_entry.set_icon_from_icon_name(gtk.EntryIconPosition.SECONDARY,
                                                       None)
-        self.find_facts()
+        self.find_facts(scroll_to_top=True)
 
     def on_search_icon_press(self, entry, position, event):
         if position == gtk.EntryIconPosition.SECONDARY:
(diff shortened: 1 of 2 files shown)


matthijskooijman:
At first glance, this fix looks good. But I tested this by scrolling down and then adding a new fact, which seems to also reset the scroll. Is that intended?

Looking at the code, you added a `reset` parameter to `FactTree.set_facts`, but it seems that is only called from `Overview.find_facts`, which always passes `reset=True`? Should maybe `find_facts()` also have a `reset` parameter, which should be `True` everywhere except when called by `on_facts_changed`? Or is this the "WIP" part of this PR? I just tried this locally and it seems to work.

Also, maybe `reset` should be `reset_scroll` or `scroll_to_top` to be more explicit?

rhertzog:
Yeah, I noticed something was fishy yesterday, that's why I moved it back to WIP until I could investigate what's wrong. I'll try out your suggested change and update the PR.

rhertzog:
So here's the updated commit which also adds `scroll_to_top` to find_facts().

But contrary to what you suggested, I only passed scroll_to_top=True in two locations: in `on_range_selected` and in `on_search_changed`.

So `on_facts_changed`, `on_row_delete_called` and `on_timeout` (and `__init__`) are not resetting the position. The `on_timeout` was the most important one to not update since it's the one that gets triggered every 60 seconds to update the figure for the currently running activity...

I believe this is now ready to be merged.

matthijskooijman (approved):
> So on_facts_changed, on_row_delete_called and on_timeout (and __init__) are not resetting the position. The on_timeout was the most important one to not update since it's the one that gets triggered every 60 seconds to update the figure for the currently running activity...

I wasn't sure about when delete triggered, I now see it happens when you delete a fact with the delete button, so makes sens to not reset the scroll.

As for timeout, I missed that one, agreed that it definitely should *not* reset the scrolling :-)

With these changes, I agree this can be merged. @mwilck, agreed?

GeraldJansen:
This merits a comment in NEWS.md!  Something like
```
## Upcoming changes in 3.0.2 (WIP)
* Fixed position in activity list upon refresh (issue 594)
```

matthijskooijman:
> This merits a comment in NEWS.md! Something like

Yeah, I was thinking to just generate NEWS entries for all relevant merges on release, but having one prepared is a good idea. As for the actual entry, I'd make it a bit more explicit:

```
* Do not reset scroll position of the activity list when the list is periodically refreshed. (issue 594)
```

rhertzog:
Yeah, I find it better to generate them at release time otherwise you generate conflicts on open merge requests for no real gain. Maybe we could add the text in the commit message so that it's just copy & paste, or could even be scripted. I'm not changing anything to the MR right now but let me know if you want me to change something related to this.

matthijskooijman:
> Yeah, I find it better to generate them at release time otherwise you generate conflicts on open merge requests for no real gain. Maybe we could add the text in the commit message so that it's just copy & paste, or could even be scripted.

See #607 for previous discussion of exactly this.

matthijskooijman:
Lacking further response from other maintainers, I've gone ahead and merged this.

PR #649, opened 2020-11-23
Add keyword entry to .desktop file

### LINKED ISSUE #629: The .desktop file misses a "keyword" entry
This is something that was flagged by the Debian package checker, lintian:

    desktop-entry-lacks-keywords-entry usr/share/applications/org.gnome.Hamster.GUI.desktop

Apparently that is a mandatory field, so would be good to supply it.

GeraldJansen:
The Debian package checker, lintian, complains about the .desktop file
missing a "keyword" entry.

Fixes #629


matthijskooijman (approved):
Looks good to me.

I'll leave this open for a few days, in case others have feedback, but lacking further response, I'll merge this without further review later this or next week or so.

rhertzog:
Looks good to me too.

matthijskooijman:
A little later than "next week", but let's merge it now :-)

Issue #636, opened 2020-10-20
Weird display of dark grey lines when filtering activities


u451f:
tags: UX, enhancement

When I filter my activities, it looks like other activities are shown as dark grey lines.
See screenshot below:
![Bildschirmfoto von 2020-10-20 17-58-24](https://user-images.githubusercontent.com/6314248/96612301-e3a1f500-12fd-11eb-92f0-9e8ef18f9167.png)

I find that this makes it a bit hard to read my screen, and see what I really did do. It also makes the scroll bar on the right extremely long.

u451f:
Oh, this was already fixed, see #615

matthijskooijman:
Nope, wasn't fixed yet (but is about to), so I'll reopen this for just a little bit :-)

PR #650, opened 2020-11-23
Remove dark bars for empty days in overview (take 2)

GeraldJansen:
The overview currently has dark bars for days with no recorded facts. These are considered unsightly by some (eg. #615, and me too). The behaviour is inconsistent because empty days at the start and end of the date range are not highlighted whereas intermediate empty days are.

This PR completely skips over empty days in the selected date range.

(This is a revival of #619, already approved by mwilk. I promise to be more patient this time.)

### DIFF
--- a/src/hamster/widgets/facttree.py
+++ b/src/hamster/widgets/facttree.py
@@ -514,7 +514,8 @@ class FactTree(graphics.Scene, gtk.Scrollable):
         days = []
         for i in range((end - start).days + 1):
             current_date = start + dt.timedelta(days=i)
-            days.append((current_date, by_date[current_date]))
+            if current_date in by_date:
+                days.append((current_date, by_date[current_date]))
 
         self.days = days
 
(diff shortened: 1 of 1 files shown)


matthijskooijman (approved):
> (This is a revival of #619, already approved by mwilk. I promise to be more patient this time.)

That made smile (in a very friendly way). Thanks, also for your other recent contributions!

The PR looks good to me!

I see that this PR contains the exact same commit as #619, so I'm assuming @mwilck's approval is still valid ;-) Together with mine, that makes two, so I think this can be merged.

Regarding the PR and commit itself, a few nitpick remarks (just for future reference):
 - You now updated the PR by merging master into your branch, but it usually gives a cleaner history when you rebase on top of master (that prevents an extra merge commit and when there are conflicts, you can edit them in the original commits rather than in the merge).
 - It is customary for commit messages to be hardwrapped (at 80 chars or so) rather than having very long lines. Not a very hard rule, but this makes reading e.g. git log output a bit easier to read.

I think I can fix these easily during the merge by doing a squash merge (since it's just a single real commit anyway). I'll also add a reference that this fixes #636, which was reported (and also closed and just reopened) after you created this commit :-)

GeraldJansen:
Thanks @matthijskooijman, also for the tips to improve the contribution process.

PR #652, opened 2020-11-26
Fix tag popup displaying by setting parent.

karliss:
Similar to other recent wayland popup issues.

Without setting parent the tag popup opens and immediately closes. Previously it only meant that popup doesn't stay opened, but recently it started interfering with focus making it difficult to type in text.

Tested on ArchLinux with GTK 3.24, Gnome desktop and wayland.

### DIFF
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -79,8 +79,7 @@ class CustomFactController(Controller):
 
         self.start_time = widgets.TimeInput(parent=self.get_widget("start time box"))
 
-        self.tags_entry = widgets.TagsEntry()
-        self.get_widget("tags box").add(self.tags_entry)
+        self.tags_entry = widgets.TagsEntry(parent=self.get_widget("tags box"))
 
         self.save_button = self.get_widget("save_button")
 
--- a/src/hamster/preferences.py
+++ b/src/hamster/preferences.py
@@ -133,8 +133,7 @@ class PreferencesEditor(Controller):
             (selection, selection.connect('changed', self.category_changed_cb, self.category_store))
         ])
 
-        self.day_start = widgets.TimeInput(dt.time(5,30))
-        self.get_widget("day_start_placeholder").add(self.day_start)
+        self.day_start = widgets.TimeInput(dt.time(5,30), parent=self.get_widget("day_start_placeholder"))
 
         self.load_config()
 
--- a/src/hamster/widgets/activityentry.py
+++ b/src/hamster/widgets/activityentry.py
@@ -203,7 +203,7 @@ class CompleteTree(graphics.Scene):
 
 
 class CmdLineEntry(gtk.Entry):
-    def __init__(self, updating=True, **kwargs):
+    def __init__(self, **kwargs):
         gtk.Entry.__init__(self, **kwargs)
 
         # default day for times without date
--- a/src/hamster/widgets/activityentry.py
+++ b/src/hamster/widgets/activityentry.py
@@ -203,8 +203,8 @@ class CompleteTree(graphics.Scene):
 
 
 class CmdLineEntry(gtk.Entry):
-    def __init__(self, **kwargs):
-        gtk.Entry.__init__(self, **kwargs)
+    def __init__(self, *, parent, **kwargs):
+        gtk.Entry.__init__(self, parent=parent, **kwargs)
 
         # default day for times without date
         self.default_day = None
--- a/src/hamster/widgets/tags.py
+++ b/src/hamster/widgets/tags.py
@@ -32,13 +32,16 @@ class TagsEntry(gtk.Entry):
         'tags-selected': (gobject.SIGNAL_RUN_LAST, gobject.TYPE_NONE, ()),
     }
 
-    def __init__(self):
-        gtk.Entry.__init__(self)
+    def __init__(self, *, parent):
+        gtk.Entry.__init__(self, parent=parent)
         self.ac_tags = None  # "autocomplete" tags
         self.filter = None # currently applied filter string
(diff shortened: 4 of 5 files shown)


matthijskooijman (commented):
Thanks for this PR. At first glance, it looks like an improvement. I haven't tested it yet, but I do have some questions about the code:
 - It seems this makes two changes: Passing a parent to the `gtk.Entry` constructor, and setting `attached_to` and `transient_for` on the  popup. Is this correct? The commit message only talks about setting a parent, so I'm slightly confused. Maybe this should be split into two commits, but at least the commit message should be a bit more explicit about what this changes and why.
 - You now pass a `parent` to the `gtk.Entry` constructor, which apparently also adds the new `Entry` to that parent? I couldn't find any docs about this, [this reference](https://developer.gnome.org/pygtk/stable/class-gtkentry.html) does not mention the `parent` parameter at all. Do you know where I should look?

Thanks!

GeraldJansen:
Probably irrelevant, but just out of curiosity I have tested this PR under Xorg and can at least report that it doesn't cause problems there.

karliss:
@matthijskooijman I am not a GTK expert so changes were done largely looking at how other entries with popups in hamster are implemented.

Considering me referring to "setting parent" as referring to whole UI element tree not the specific property or method containing word parent. The most important thing is correctly attaching the popup, to the window and or entry input element. But unless I also parented the Entry to it's container GTK was complaining about popup being an unparented top level window and immediately disappearing.  Probably due to `self.get_ancestor(gtk.Window)` working only after self which is Entry being placed in UI tree, which makes sense. 

From practical observations to fix the main problem only setting parent for `Entry` and `set_transient_for`. `set_attached_to` documentation talks about "logical binding between the window and the widget" and style property propagation. With "Gtk.Menu created by a Gtk.ComboBox, a completion popup window created by Gtk.Entry" given as examples for where it's useful, and other widgets in Hamster with similar completion popups do the same so it seemed like the proper way for implementing such popups.

> I couldn't find any docs about this, this reference does not mention the parent parameter at all. Do you know where I should look?

The reference you linked is for PyGtk 2.0 and 2.24 that's from what I understand that's GTK2 python wrappers. I am not surprised that there are some differences. Something like https://www.gtk.org/docs/language-bindings/python/ and https://lazka.github.io/pgi-docs/#Gtk-3.0 should be more up to date . Although it won't help too much in this specific case as it documents most  constructor arguments as `**kwargs` . Which basically means there are too many arguments, most of them forwarded to common base class and we can't be bothered to properly document them everywhere, look at the examples or read the source code. This problem affects even parts of Python standard library.

matthijskooijman:
@karliss thanks for your additional explanations, it all makes sense now :-)

> But unless I also parented the Entry to it's container GTK was complaining about popup being an unparented top level window and immediately disappearing. Probably due to self.get_ancestor(gtk.Window) working only after self which is Entry being placed in UI tree, which makes sense.

Ah, that explains why adding `parent` to the constructor arguments (instead of adding the `TagEntry` to its parent manually *after* the constructor completes, which was the old behavior) is needed, makes sense. If I omit this part, ` self.get_ancestor(gtk.Window)` indeed returns `None`. It seems passing `parent` to the constructor happens for `CmdLineEntry` already, probably for the same reason.

Looking at this more closely, I found that `TimeInput` also uses `get_ancestor()` but does not pass the `parent` to the constructor when used in the preferences window, so I'd guess that if you open up the preferences window and click the time input, that popup also doesn't work properly on Wayland.

I've fixed this in an additional commit, and changed the constructors of `TimeInput`, `CmdLineEntry` and `TagEntry` to make the `parent` argument a required and keyword-only argument (to prevent this issue from being accidentally reintroduced).

I've pushed these changes to your branch, updating this PR. @karliss, could you see if that looks ok and still works well for you? I've also updated the commit message to include part of your explanations above. Note that I force pushed to reorder commits and update the commit message, so pull carefully.

> and we can't be bothered to properly document them everywhere,

Right, then we'll just assume that the `parent` argument does what you'd expect from it. It seems it's used like this elsewhere too and it seems to hold up so far, so that's fine even without docs :-p

> Probably irrelevant, but just out of curiosity I have tested this PR under Xorg and can at least report that it doesn't cause problems there.

@GeraldJansen Confirming that it still works on Xorg seems quite relevant to me, so thanks for testing. If you want to give this updated PR another try, feel free (but I already tested on Xorg, so it's at least not completely broken :-p).

matthijskooijman (approved):

matthijskooijman:
I rebased this on top of the current master, after merging a few PRs, for a slightly more up-to-date test.

GeraldJansen:
@matthijskooijman I have taken this for another spin on X11 and all the popups and popovers still seem to be okay.

karliss:
Tested, cmdline, time entries and tags are fine.

 Noticed that for category and activity popup position is wrong, but those seem to be using builtin GtkEntryCompletion so I dont think there is anything we can improve for them right now. The amount of position error changed based on current monitor. It was fine on leftmost monitor but offset to the right on rest of them.

matthijskooijman:
> Noticed that for category and activity popup position is wrong, but those seem to be using builtin GtkEntryCompletion so I dont think there is anything we can improve for them right now. The amount of position error changed based on current monitor. It was fine on leftmost monitor but offset to the right on rest of them.

Hm, interesting. Could you share a screenshot?

matthijskooijman:
Given positive testing from @karliss  and @GeraldJansen, I'm going to merge this. @karliss would you open a new issue for the mispositioning of those popups with a screenshot? Even if there's nothing obvious we can fix right away, it would be good to have an issue to document the issue for later.

PR #661, opened 2020-12-22
Fix documentation in Download & Unzip section

aayushagra:
The documentation for the edited section is missing the unzip command and gets the name of the output file wrong. This fixes both.


benjaoming (approved):
Thanks, that's a nice catch!

Issue #657, opened 2020-12-11
Inputting descriptions with single or double comma?


matthijskooijman:
From https://github.com/projecthamster/hamster-shell-extension/issues/334:

> With hamster 3.0.2, if you type Activity@Category, task description in the shell extension, nothing happens. To get it to actually start, you need to type Activity@Category,, task description. This is likely related to the breaking change made in hamster that now requires two commas to separate descriptions:
projecthamster/hamster@42d3c05

> I think we should revisit that decision as it doesn't seem like a good decision to me. In particular when typing Activity@Category, #tag just works.

> But if we don't, then we somehow need to fix the shell extension. Something needs to happen when the user inputs something... either it creates an activity with some unexpected values, or it displays an error. But just dropping the user's input is wrong.,

And from https://github.com/projecthamster/hamster/pull/656#issuecomment-743337886

> I would love to see some discussion on projecthamster/hamster-shell-extension#334 and see wether we prefer to revert 42d3c05 here or find some other solution elsewhere.

matthijskooijman:
w00ps, pressed ok too soon. Edited in the issue description above now :-)

As for the double comma, I also noticed that this had some special handling in the activity editing window that seems a little fragile. It seems that the description is also the only part that is not serialized into the cmdline, though it is editable from the commandline using the double comma I think.

Anyway, I wonder what the motivation for the double comma is. I guess it serves to distinguish a tag from the description? Maybe it allows starting the description with a hash, which is probably not otherwise possible? Would we mind?

I haven't looked too closely at this yet, though, just wanted to get this dicussion in a separate issue in the right place :-)

GeraldJansen:
Basically the double comma emerged to satisfy the request of one user to be able to include a comma in the activity name. @mwilck argued against the change [here](https://github.com/projecthamster/hamster/issues/465#issuecomment-547335487) and I have to say I agreed with him then and still do.

ederag:
> Basically the double comma emerged to satisfy the request of one user to be able to include a comma in the activity name.

You do not think this through. https://github.com/projecthamster/hamster/issues/465#issuecomment-557810768 + other reasons.

GeraldJansen:
@ederag The fact that I didn't provide further opposition to the double comma doesn't mean that I liked it. In fact, I agree that the decision could and should be revisited. It could be done without reducing the rock solidness of the parser.

mwilck:
[...]

PR #662, opened 2020-12-23
Revert to single comma for description barrier

GeraldJansen:
This PR reverts the use of a double comma to indicate the start of the description, a breaking change introduced in Hamster 3.0, back to the previous use of a single comma, as discussed in #657. Likewise, the double comma needed before tags, in the case of descriptions containing the #hash pattern, is also reverted to the use of a single comma.

This change requires two restrictions to the parsing rules. Firstly, no comma is allowed in the activity name (#280). Secondly, tags
may not be separated by a comma when entered on the commandline (ie. just use `#tag1 #tag2`, not `#tag1, #tag2`). The differences in tests/test_stuff.py illustrate the changes in behaviour of the parser.

### DIFF
--- a/help/C/input.page
+++ b/help/C/input.page
@@ -11,18 +11,18 @@
     To start tracking, press the <gui style="button">+</gui> button,
     type in the activity name in the entry,
     and hit the <key>Enter</key> key.
-    To specify more detail on the fly, use this syntax: 
-    `time_info activity name @category,, some description #tag #other tag with spaces`.
+    To specify more detail on the fly, use this syntax:
+    `time_info activity name@category, some description #tag #other tag with spaces`.
 </p>
 
 <steps>
     <item><p>Specify specific times as `13:10-13:45`, and "started 5 minutes ago" as `-5`.</p></item>
     <item><p>Next comes the activity name</p></item>
     <item><p>Place the category after the activity name, and start it with an at sign `@`, e.g. `@garden`</p></item>
-    <item><p>If you want to add a description, add a double comma `,,`.</p></item>
-    <item><p>The description is just freeform text immediately after the double comma, and runs until the end of the string or until the beginning of tags.</p></item>
+    <item><p>If you want to add a description, add a comma `,`.</p></item>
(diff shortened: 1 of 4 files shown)


rhertzog:
Thanks for this work, someone had to tackle it. But I find the restriction to not have to use a single comma between two tags strange. What happens if you do?

Also I'm wondering why we are not extracting tags out of the description when it contains hash keywords. Sometimes my description looks like `review #bugs in #hamster` and I would like the description to be turned into `review bugs in hamster, #bugs #hamster`. I don't think we should allow spaces in tags.

GeraldJansen:
@rhertzog Tags are searched from the end of the line backwards, stopping at the first comma. With my changes, if you separate tags by commas, all but the last will be considered part of the description. 

I tried to make the minimal changes necessary to revert from the use of double commas back to the use of a single comma. These changes continue to allow one to include spaces and in fact any character other than comma and '#' within tags. Allowing spaces within tags further complicates parsing, and I personally would also disallow that, but that would be going another step further.

In my opinion, your request to parse the description field for additional tags should be considered as a new feature request, separate from the issue of reverting from a double comma back to a single comma. I created the wiki page  [Entering-activities](https://github.com/projecthamster/hamster/wiki/Entering-activities) in attempt to define how the parser should behave outside of the code and test_stuff.py file, but in the end that discussion was never finalized. At any rate, I feel that discussion goes beyond the scope of this PR. (Better is the enemy of good.)

rhertzog:
I understand the desire to keep things separate but this has been irking me for a long time so I went ahead and submitted #663 which builds on top of your work to simplify how tags are parsed and to implement this feature.

ederag:
The wiki page is outdated, and should have been removed after v3 was out.
Here are the correct v3 instructions: https://github.com/projecthamster/hamster/issues/657#issuecomment-751372519

GeraldJansen:
> The wiki page is outdated, and should have been removed after v3 was out.

@ederag. Okay, I deleted it.

GeraldJansen:
Closed in favour of #663.

PR #665, opened 2020-12-26
read 'checked' status correctly and clean up variable names and dead code

### LINKED ISSUE #664: html export checkboxes don't work
If I export an html report and view it in Chromium 83 on Debian stable KDE, nothing happens when I uncheck the `activites`, `categories` or `tags` checkboxes.
I observe that changing the js from 'attr' to 'prop' fixes the issue, like:

```
--- a/data/report_template.html
+++ b/data/report_template.html
@@ -297,13 +297,13 @@
 
             // determine if we will be doing any internal counting
             var keys = [];
-            if ($("#show_activities").attr("checked"))
+            if ($("#show_activities").prop("checked"))
                 keys.push("activity");
 
-            if ($("#show_categories").attr("checked"))
+            if ($("#show_categories").prop("checked"))
                 keys.push("category");
 
-            if ($("#show_tags").attr("checked"))
+            if ($("#show_tags").prop("checked"))
                 keys.push("tags");
```

drewbenn:
Fixes #664

### DIFF
--- a/data/report_template.html
+++ b/data/report_template.html
@@ -260,25 +260,6 @@
                 return false;
             })
 
-            $("#show_details").click(function() {
-                if ($(this).attr("checked")) {
-                    $(".by_date_row").show();
-                } else {
-                    $(".by_date_row").hide();
-                }
-
-            })
-
-            $("#show_totals").click(function() {
-                if ($(this).attr("checked")) {
-                    $(".by_date_total_row").show();
-                } else {
-                    $(".by_date_total_row").hide();
-                }
-
-
-            })
-
             populateFactTable();
 
 
@@ -297,13 +278,13 @@
 
             // determine if we will be doing any internal counting
             var keys = [];
-            if ($("#show_activities").attr("checked"))
+            if ($("#show_activities")[0].checked)
                 keys.push("activity");
 
-            if ($("#show_categories").attr("checked"))
+            if ($("#show_categories")[0].checked)
                 keys.push("category");
 
-            if ($("#show_tags").attr("checked"))
+            if ($("#show_tags")[0].checked)
                 keys.push("tags");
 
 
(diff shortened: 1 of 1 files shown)


GeraldJansen:
Nice. Works for me with both Chromium and Firefox.

data/report_template.html
```diff
@@ -297,13 +278,13 @@
 
             // determine if we will be doing any internal counting
             var keys = [];
-            if ($("#show_activities").attr("checked"))
+            if (document.getElementById("show_activities").checked)
```
rhertzog, line 281: I believe you should keep the jquery syntax but use the `.prop("checked")` function or `.checked` attribute.
```suggestion
            if ($("#show_activities").checked))
```
Same goes for the two other occurrences. Not tested but it should work AFAIK.

rhertzog: Actually I did test my change now and it doesn't work... not sure why though. I was pretty convinced that I was correct after having read https://api.jquery.com/prop/

Apparently it's been a long time that I haven't used jQuery... I notice though that we have a very old jQuery (1.4.1 vs 3.5.1), maybe it's partly related.

rhertzog:
```suggestion
            if ($("#show_activities")[0].checked)
```

This one is tested and works!

matthijskooijman:
This looks like a good change to me, but I agree with @rhertzog that it would be good to keep jquery syntax. @drewbenn, could you update your PR to include Raphaël's suggestions?

Ideally, IMHO, this PR would be 3 separate commits (fix checked, remove dead code, fix `for` attributes), but given the entire thing is simple enough and the commit messages lists all changes, I don't think this should block merging at all.

GeraldJansen:
@matthijskooijman I strongly doubt @drewbenn would object to you just forcing Raphaël's suggestions and merging. (Aside: Let's encourage contributions by avoiding "better is the enemy of good".)

Issue #667, opened 2021-01-16
Calendar date picker date mismatch


kidorion:
The date picked in graphical UI calendar doesn't match the date applied to command line.
My guess is a locale problem. My locale is Sydney, Australia = UTC+10 or UTC+11 during summer.
I'm using Linux Mint Debian 10.2 buster, Cinnamon desktop 4.6.7.
Goal: try and make an entry for yesterday Friday 2021-01-15 09:00-15:00
Thanks..
Steps to reproduce are:
1. Add activity
1. Enter start time as 09:00
1. Select started date in calendar graphical UI as 2021-01-15
1. Date in command line is 2021-01-14
See image:
![image](https://user-images.githubusercontent.com/53280834/104797038-5cd2ba80-580f-11eb-924b-d90941472377.png)

Also browsing past entries using the Graphical UI show similar out-by-one-day errors.

GeraldJansen:
Hamster just works with local times so I don't believe your location is the issue. More likely this is a duplicate of previously reported issues being worked on in #630. (Hopefully work on this very important issue will be resumed soon!)

ederag:
There has been a fix pending for 8 months and a half: https://github.com/projecthamster/hamster/pull/597#issuecomment-622463737

Pops34:
same problem
It's the 17th, I want to add an activity from the previous day, the 16th, I go back to Tuesday the 16th, the software displays Monday the 15th (inconsistent date), and when I want to correct the date by adding an activity, there is a difference between the selected date and the saved command line.  
![bug1](https://user-images.githubusercontent.com/62051042/108162200-e87c8700-70ec-11eb-9c54-f788e065baa7.png)
![bug2](https://user-images.githubusercontent.com/62051042/108162202-e9151d80-70ec-11eb-821b-95a87306ab5f.png)

ederag:
A fix has been pending for 9 months and a half: https://github.com/projecthamster/hamster/pull/597#issuecomment-622463737

GeraldJansen:
@ederag That fix has been rejected by current maintainers and closed for several months now. If you are still interested in making a positive contribution to the project please consider making a new minimal pull request based on current master. Otherwise please do us all a favour and move on.

ederag:
> do us all a favour and move on

Interesting tone.
Where is your aspiration to democracy (I just give information to this community) ? 
Or was it just about power ?
Do you realize that the current maintainers, by letting this issue rot, and letting duplicates pile up, 
[...]

PR #670, opened 2021-02-16
hamster/client: edit_activity: don't touch self.date on date changes

lynxis:
The fact is recalculated already in on_start_date_change(). By calling the setter of
self.date the fact would be change a second time. The start_date was then wrong by
the delta. E.g. edit a complete fact (start and end must be set). Click on the start date and choose 2 days into the future, the command line shows now 4 days into the future.

Fixes: #667

### DIFF
--- a/src/hamster/edit_activity.py
+++ b/src/hamster/edit_activity.py
@@ -248,7 +248,6 @@ class CustomFactController(Controller):
                     # preserve fact duration
                     self.fact.end_time += delta
                     self.end_date.date = self.fact.end_time
-            self.date = self.fact.date or dt.hday.today()
             self.validate_fields()
             self.update_cmdline()
 


ederag:
Better fix, been pending for 9 months and a half: https://github.com/projecthamster/hamster/pull/597#issuecomment-622463737

benjaoming (approved):
Thanks for the change and the fix - LGTM

matthijskooijman:
I haven't looked closely at this PR yet, but it seems very simple and I recall (from earlier discussion and fix attempt in #630) that this code is quite complex with a lot of complex interactions. So I'm afraid that this simple fix might have unintended side effects.

@benjaoming Did you also test this PR, or just reviewed the code?

benjaoming:
No, maybe @lynxis can reflect on the extend of how this is tested?

Could letting a small change like this in to the `master` branch also provides a way forward for verification before a release?

GeraldJansen:
I have tested this simple change. Tests in tests/test_stuff.py all pass. The coordination of cmdline and date pickers seems to work perfectly with this change. The left and right arrows also decrement and increment the dates properly in the cmdline and date pickers. Activities straddling the hamster start of day also seem okay. The only glitch I've found is that the day displayed on the timeline (hamsterday) doesn't get updated if the date is changed in the cmdline or start date picker, so the timeline display gets out of sync. This in turn affects whether or not the start date is displayed on the command line. Anyway, complements for the dramatic improvement in ability to edit facts on prior dates with such a minimal change.

lynxis:
I've tested only a couple things in the edit_activity and see the update works correct.

- click on the start date change it to a different date as in the main window (hamsterdate?).
- click on the end date and see it's matching the cmdline
- change the start_date back to the original date. the start_date in the cmdline got removed (no idea what's intended)

GeraldJansen:
> * change the start_date back to the original date. the start_date in the cmdline got removed (no idea what's intended)

The start_date is not displayed in the cmdline if it is the same as hamster day.

ederag:
> The start_date is not displayed in the cmdline if it is the same as hamster day.

One needs to be more precise, see https://github.com/projecthamster/hamster/pull/630#issuecomment-731761920:
The timeline shows a _hamster day_ (the default day)
while the start date is a _civil date_.
Saying they are "the same" makes little sense.

The rule:
https://github.com/projecthamster/hamster/blob/0ba8407b94227a694f6b635dece79f2aef3059c4/src/hamster/lib/datetime.py#L439-L441

lynxis:
Anything missing for a merge?

benjaoming:
Can someone say with confidence that it's an overall improvement? In that case, I think we should merge it :)

It _sounds_ like it's an overall improvement :)

ederag:
Had the fix been that simple, it would have been proposed earlier.
The [timeline display getting out of sync](https://github.com/projecthamster/hamster/pull/670#issuecomment-780684409) issue mentioned by @GeraldJansen is important.
Besides, merging this PR would prevent easy merging of my fix, 
which to the very least is a much better start
(it fixes both the core issue, and clarifies code for future maintenance), 
see https://github.com/projecthamster/hamster/pull/597#issuecomment-780524285.

matthijskooijman:
I had another look at this issue. I considered (and even started somewhat) working on a more thorough refactor/cleanup that would fix this issue, but then realized that that would probably take quite some time to make and then also to get reviewed and merged, so that a small fix like in this PR would probably be better now, delaying the more thorough cleanup to later.

Looking at this PR again, I realized that the problem is that the fact date is updated twice (once by `on_start_date_changed()` and one by the `date` setter) and this PR solves this by removing the call to the `date` setter from `on_start_date_changed()`. However, this has the side effect of no longer setting the `date` property and no longer updating the dayline, as @GeraldJansen found in testing as well. An obvious alternative fix would then be to modify the `date` setter to no longer update the fact date, which would also prevent that from happening twice. I considered and tried this alternative, and it seems to work very well, without any additional side effects.

The result is in a new PR, #674. @lynxis, @GeraldJansen, @benjaoming, would you care to have a look at that PR and let me know what you think?

matthijskooijman:
I've just merged #674, which fixes this issue in a slightly different manner, so I'm closing this one. @lynxis Thanks again for your effort, which provided the stepping stone for that fix :-)

Issue #595, opened 2020-04-09
Resume using Ctrl-Space (v3)


otto-dev:
Feature request: In `v3.0.1` it possible to stop tracking the current task using `Ctrl-Space`. It would be nice if pressing `Ctrl-Space` at a later time would recreate a new entry for the most recent task, tracking from the current time. This would allow toggling the tracking of a task both on and off.

This would substitute the functionality in v2, where it was possible to start/resume tracking a task, starting at the current time, by double-clicking a previous existing entry.

PR #620, opened 2020-06-14
Resume last activity using Ctl-Space.

GeraldJansen:
This allows toggling the tracking of an activity on and off with Ctl-Space, as requested in #595.

PR #678, opened 2021-04-25
Resume last activity using Ctl-Space.

GeraldJansen:
This allows toggling the tracking of an activity on and off with
Ctl-Space, as requested in #595.

### DIFF
--- a/help/C/keys.page
+++ b/help/C/keys.page
@@ -34,15 +34,15 @@ than one page.</p>
 All details of the selected activity, except start/end, are copied to the new activity.</p>
 
 <p><key>Ctrl</key>+<key>+</key>: Same as the <key>+</key> button.
-Create a new activity. 
-If an existing activity is selected, 
-it will be resumed (same as <key>Ctrl</key>+<key>R</key>), 
+Create a new activity.
+If an existing activity is selected,
+it will be resumed (same as <key>Ctrl</key>+<key>R</key>),
 otherwise, new from stratch (same as <key>Ctrl</key>+<key>N</key>).</p>
 
 
 
 <p><key>Ctrl</key>+<key>Space</key>: Stop tracking the ongoing activity, if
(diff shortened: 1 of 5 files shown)


GeraldJansen:
@matthijskooijman This is an identical replacement for #620. There were problems with that PR because last year I deleted my github fork out of impatience. Sorry for the confusion.

GeraldJansen:
I have just tested it again and am fairly confident it works as advertised. I also tested xfce4-hamster-plugin (current master) against this PR just to be sure I didn't muck up dbus functions. Nothing seems broken. Pinging @sh-zam who liked the earlier PR and @otto-dev who requested the feature to see if they did any testing.

sh-zam:
> I have just tested it again and am fairly confident it works as advertised. I also tested xfce4-hamster-plugin (current master) against this PR just to be sure I didn't muck up dbus functions. Nothing seems broken. Pinging @sh-zam who liked the earlier PR and @otto-dev who requested the feature to see if they did any testing.

I've cherry-picked the previous patch locally and have been using it ever since it was created. So, yes it works just fine :)

GeraldJansen:
Ping: anything holding up a merge on this one?

matthijskooijman:
Good point, given two maintainers are involved, I think this can indeed be merged.

I did a quick test just now, seems to work as expected. One thing that was a bit surprising is that if you stop an activity and restart it immediately with ctrl+space, it creates a *new* fact that follows the previous fact directly (start time = stop time of previous fact). It seems I would have expected that it would instead continue the previous fact (I think this is what the gnome shell extension does when you stop and immediately restart the same activitity). However, given that this is really mostly a pathologic case (in real usecases you will have some time between stopping and restarting the activity, unless you accidentally stopped and want to resume), I don't think this is really important. So I'm going to go ahead and merge this, and we can always improve on this later if we think it is relevant enough.

PR #696, opened 2022-01-20
refactoring code with List Comprehension

zjzh:
refactoring code with List Comprehension which is more pythonic, concise and efficient; how do you think this change which has practical value？


mwilck:
Waf is just a tool for us. We don't need style patches for it.

PR #699, opened 2022-01-29
Fix time-activity overlap with wide fonts

### LINKED ISSUE #698: Time in Activity Overview Overlaps Activity Title
Greetings all,

On GNOME 3.38 I have noticed an overlap in the characters delineating the date and the title of an activity. 
```
OS: Debian 11 bullseye
Kernel: x86_64 Linux 5.10.0-11-amd64
Shell: bash 5.1.4
DE: GNOME 3.38.4
Font: Noto Sans Mono 12
WM: Mutter
```
Couldn't find an open issue to follow.  It appears that the date column width is statically set and does not scale to wider fonts such as many moon space type fonts used in other distributions. 

```
Hamster Version: 3.0.2
```

image

GeraldJansen:
Fixes #698. The width alotted to the time interval in the overview was
fixed at 100 pixels but that was not sufficient when using large/wide
fonts. This PR uses the pixel width of ""00:00 - 00:00" when rendered
using the current default font.

### DIFF
--- a/src/hamster/widgets/facttree.py
+++ b/src/hamster/widgets/facttree.py
@@ -76,7 +76,7 @@ class Label(object):
         self._label_context = cairo.Context(cairo.ImageSurface(cairo.FORMAT_A1, 0, 0))
         self.layout = pangocairo.create_layout(self._label_context)
         self.layout.set_font_description(pango.FontDescription(graphics._font_desc))
-        self.set_text("Hamster")  # dummy
+        self.set_text("00:00 - 00:00")  # dummy time_label for finding width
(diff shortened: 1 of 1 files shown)


aquaherd (approved):
If there's no veto I am going to merge this.

matthijskooijman (approved):

aquaherd:
While testing this I put in a 20pt Font and it looked good on the time ranges. 

However in the German locale, the weekday names were cut off. 

Maybe you could apply the same approach to calculate the leftmost panel width?

PR #701, opened 2022-04-13
Make hamster-cli.py and hamster-windows-service.py executable

matthijskooijman:
This are intended to be executed as binaries and already have
a hashbang, so it makes sense to have them executable. Installation
already handles this, but having them executable in the repo is
convenient for debugging.

hamster-service.py was already executable, this seems to have been
accidentally commit in commit 19d16009 (Resume last activity using
Ctl-Space).


mwilck (approved):
Looks ok.

PR #702, opened 2022-04-13
Fix facttree touchpad scrolling on Wayland by removing custom scroll code

matthijskooijman:
This PR was prompted by touchpad scrolling not working on wayland, because this is handled using different ("smooth") scroll events that were not handled by the custom scrolling code.

This is fixed by removing all the custom scrolling code, making the facttree DrawingArea just big enough to fit all facts and just letting gtk handle scrolling by moving the facttree inside a viewport. By only redrawing dirty areas, invisible facts are (still) not drawn (and it seems that when scrolling, there is now less need to redraw previously drawn areas as well).

Leading up to this change are a few more commits with additional cleanups, mostly removing unused code or small refactorings that make the main change easier to implement.

There might be subtle changes (i.e. in how far things are scrolled when pressing pgdown), but there should be nothing that is actually significant.

Tested on Wayland and Xorg (both inside Gnome), seems to work as expected.

### DIFF
--- a/src/hamster/overview.py
+++ b/src/hamster/overview.py
@@ -559,7 +559,7 @@ class Overview(Controller):
     def on_stop_clicked(self, button):
         self.storage.stop_tracking()
 
-    def on_row_activated(self, tree, day, fact):
+    def on_row_activated(self, tree, fact):
         self.present_fact_controller("edit", fact_id=fact.id)
 
     def on_row_delete_called(self, tree, fact):
--- a/src/hamster/widgets/facttree.py
+++ b/src/hamster/widgets/facttree.py
@@ -33,20 +33,6 @@ from hamster.lib import stuff
 from hamster.lib.fact import Fact
 
 
-class ActionRow(graphics.Sprite):
-    def __init__(self):
-        graphics.Sprite.__init__(self)
-        self.visible = False
-
-        self.restart = graphics.Icon("view-refresh-symbolic", size=18,
-                                     interactive=True,
-                                     mouse_cursor=gdk.CursorType.HAND1,
-                                     y=4)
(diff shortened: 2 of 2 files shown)


mwilck (changes requested):
Tested under GNOME 42. Mouse/wheel and touchpad work nicely, but I noticed that the PgUp / PgDown keys don't work any more. They work with current master. Is this intentional?

matthijskooijman:
> Tested under GNOME 42. Mouse/wheel and touchpad work nicely, but I noticed that the PgUp / PgDown keys don't work any more. They work with current master. Is this intentional?

Ah, good catch. This worked when I tested it during development earlier, but it seems a later changes broke this. They are supposed to be handled by ScrolledWindow, not sure why they are not. I'll have a closer look soon.

GeraldJansen:
Just in the way of encouragement, I have also tested this PR under xorg+xfce4 and also found it works fine (except for PgUp/PgDown).

GeraldJansen:
Another minor imperfection (on xorg at least), is that scrolling to the bottom with either KEY_End or repeated KEY_Down, the Total line for the last day doesn't show. It does, however, when scrolling to the bottom with the mouse wheel.

matthijskooijman:
> Another minor imperfection (on xorg at least), is that scrolling to the bottom with either KEY_End or repeated KEY_Down, the Total line for the last day doesn't show. It does, however, when scrolling to the bottom with the mouse wheel. (Followup: the same is true for current master, but a final PgDown does show the last line.)

I see the same with 3.0.3, also on Wayland. It doesn't seem to be a regression in this PR, but maybe something to fix while working on that code anyway indeed.

Issue #416, opened 2019-05-19
Move to web-based documentation?


GeraldJansen:
In view of the current situation w.r.t. packaging and installation (see #399), I would like to propose moving the documentation pages towards a web-based solution like readthedocs.org. The application's current Help menu item would be replaced by an About dialog with a link to the URL of the web documentation. This would reduce the complexity of manual installation because normal users would not have to build and install the documentation at all and thus would not need dependencies such as gnome-doc-utils xlstproc docbook-utils yelp. I would be willing to work on this if there is agreement about the direction.

GeraldJansen:
For a proof of concept, see https://geraldjansen.github.io/hamster/ which is based on the markdown files here https://github.com/GeraldJansen/hamster/tree/master/docs.

ederag:
On the one hand your arguments are totally valid,
and the proof of concept looks nice at first sight.
On the other hand, mallard/yelp docs look nice and reactive,
and I'm reluctant to throw away so much work done by our predecessors,
in particular the translations.

GeraldJansen:
If we decide to proceed I will try to retain the translations. The translated pages would be viewable at https://???.github.io/hamster/xx/index.html (xx in fr, de etc.) and the link in the About dialog would point to the appropriate language specific page. 

It would also be possible to store previously released versions in version-specific subdirectories, eg. docs/v2.2/xx/*.

ederag:
Fine ! Is the page generation process automated ?

GeraldJansen:
When the markdown pages are pushed/merged to the master /docs dir, the github.io pages are updated automatically. I am not yet sure how to automate the translation process, so initially I would foresee a developer running a script to generate markdown in various languages and making a pull request if there are any changes. I don't expect much action on the doc pages so it shouldn't be too much work. There should probably be an effort to update the docs before new significant releases. My time availability is also limited so no promises on timelines.

ederag:
If a single developer has to generate web pages,
then why not use the existing infrastructure and use `yelp-build` ?
http://projectmallard.org/about/learn/tenminutes.html
Hold on. Got to think about it.

But I agree that the documentation build process is separated from the installation scope, for now.

GeraldJansen:
[...]

PR #706, opened 2022-06-22
Static html of the english help pages

RJDan:
I have made static html of the english help pages as per #416


mwilck:
Hm, why HTML and not markdown? I can see that "static HTML" is mentioned in #416, but I believe it wasn't meant that way. Writing explicit HTML seems pretty much backward these days.

RJDan:
I'd be happy to convert to md in a bit. I had no idea what static html meant. That was my best guess.

PR #707, opened 2022-06-30
Update i18n.py

piio:
Fix for python 3.11 - module.bind_textdomain_codeset was removed


aquaherd:
Will merge !715 instead

Issue #688, opened 2021-11-02
Submit the application to FlatHub as well


rabin-io:
I stared an App Submission process with flathub, you are welcome to take over anytime.

flathub/flathub#2614

rabin-io:
approved https://github.com/flathub/flathub/pull/2636
and it's available on FlatHub now, https://flathub.org/apps/details/org.gnome.Hamster

mattdrepo:
flatpak run org.gnome.Hamster
Traceback (most recent call last):
  File "/app/bin/hamster", line 40, in <module>
    from hamster import client, reports
  File "/app/lib/python3.9/site-packages/hamster/reports.py", line 33, in <module>
    from hamster.lib.configuration import runtime
  File "/app/lib/python3.9/site-packages/hamster/lib/configuration.py", line 128, in <module>
    runtime = RuntimeStore()
  File "/app/lib/python3.9/site-packages/hamster/lib/configuration.py", line 124, in __init__
    self.storage = Storage()
  File "/app/lib/python3.9/site-packages/hamster/client.py", line 79, in __init__
    self.bus = dbus.SessionBus()
  File "/app/lib/python3.9/site-packages/dbus/_dbus.py", line 212, in __new__
    return Bus.__new__(cls, Bus.TYPE_SESSION, private=private,
  File "/app/lib/python3.9/site-packages/dbus/_dbus.py", line 102, in __new__
    bus = BusConnection.__new__(subclass, bus_type, mainloop=mainloop)
  File "/app/lib/python3.9/site-packages/dbus/bus.py", line 124, in __new__
    bus = cls._new_for_bus(address_or_type, mainloop=mainloop)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NotSupported: Using X11 for dbus-daemon autolaunch was disabled at compile time, set your DBUS_SESSION_BUS_ADDRESS instead

mattdrepo:
resolved with
 export $(dbus-launch)
before launching hamster from cli

aquaherd:
May I close this issue, then?

PR #708, opened 2022-06-30
Update Flatpak installation instructions

salim-b:
Following https://github.com/projecthamster/hamster/issues/688

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Installation:
 (or graphical package installer).
 
 ##### Snap
+
 Easy installation on any distribution supporting snap:  
 https://snapcraft.io/hamster-snap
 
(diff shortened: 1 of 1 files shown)


aquaherd (approved):

matthijskooijman:
I had a look over this PR, and good to point people at flathub. I do have two remarks:
 - The instructions assume that a "flathub" remote is configured in the user's flatpak installation, but I recently tried installing something else from flathub and found that flathub was not configured by default and needed to be added manually first (this was On Ubuntu 22.10, not sure if this is universal). I would like these instructions to be useful to someone without prior knowledge about how flatpak works exactly, so if prior setup is needed, it would IMHO be good to at least mention this, and ideally also point to some official docs that document how to set this up.
 - The PR as-is replaces info on how to install a separately downloaded flatpak file with how to install from flathub, but I think it would be good to (also) document the manual install method, since it allows installing flatpak files built manually or by our github workflows (i.e. to allow easily testing git master or a pull request). Bonus points for providing concise instructions on where to find such workflow-generated flatpak builds.


Having said that - I think that these are not blockers, we can also merge this and fix those issues later.

salim-b:
> Having said that - I think that these are not blockers, we can also merge this and fix those issues later. 

Given that this PR is about a trivial documentation change but still has been open for 3/4 year now, I'd appreciate this way forward.

matthijskooijman:
Merged! I'm thinking to further improve/restructure the installation instructions a bit after 3.0.3 is released, so I can then also include the above.

PR #709, opened 2022-08-06
Croatian translation 20220806

milotype:


aquaherd:
If there's no veto, I am going to merge this.

PR #685, opened 2021-07-19
Bump flatpak runtime version and deps

tchernobog:
Bump to latest GNOME Sdk/Platform version 40.

PR #717, opened 2023-04-24
Use ubuntu 20.04 for workflows

matthijskooijman:
This fixes the flake8 workflow (which failed due to missing python 3.5 on ubuntu-latest) and fixes the test suite on python3.5 (due to no longer supported explicitly configured ubuntu-18.04). See commit messages for a bit more details.

Note that that with this change, we could also consider lowering the required python version from 3.6 back to 3.5 (it was only raised by @tchernobog in #685 in the assumption that ubuntu-18.04 did not have 3.5, but looking at the [python-versions](https://github.com/actions/python-versions/blob/main/versions-manifest.json) repo, which is used by the [setup-python action](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#available-versions-of-python-and-pypy), I believe it *is* actually available), but since python 3.6 is pretty well available (even Debian oldstable buster has 3.7), I guess it is fine to leave it at 3.6.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,10 +16,15 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Test 3.6, which is our minimum version, on 18.04 since the
-          # latest Ubuntu does not have it.
-          - os: ubuntu-18.04
-            python-version: 3.6
+          # This tests uses the oldest Ubuntu and its oldest Python
+          # version, to get as close to our minimum required version as
+          # possible. This does mean that we will not catch changes to
+          # the code that unintentionally reaise the required python
+          # version, but there's no easy way to run an older python
+          # version (without having to install all or dependencies from
+          # source) it seems.
+          - os: ubuntu-20.04
+            python-version: 3.8
           # On the latest ubuntu, just test the default 3.x version.
           - os: ubuntu-latest
             python-version: 3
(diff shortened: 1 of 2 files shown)


matthijskooijman:
This is fairly trivial change, so if it indeed fixes the build, I might merge it without further reviews.

matthijskooijman:
Hm, seems that didn't work - turns out the test suite uses apt to install python rather than the get-python action, so it does not have py3.6 available anymore. Maybe we *can* use get-python here, or otherwise just run the test suite with a newer python version...

(also, the linting still fails, but that's because there is a styling error triggered, possibly something that was added to flake8 recently).

Gotta run, will try to look closer tomorrow.

matthijskooijman:
> Maybe we can use get-python here

Nope, this is what I wrote in comments earlier:

    # This runs the test suite. Since we need some system dependencies
    # (e.g. pygobject) that cannot easily be installed inside a
    # virtualenv using pip (that requires installing a dozen -dev
    # packages and building parts of pygobject from source, which makes
    # things way too complicated), this just runs with the Ubuntu system
    # python (e.g. not using the standard setup-python action, which
    # sets up a virtualenv).

So maybe it would be possible, but too complicated.

> or otherwise just run the test suite with a newer python version...

I've just pushed an update that does this. It's annoying that we do not validate py3.6 compatibility anymore now, but it's not a disaster (and IMHO also not a reason to bump the minimum required python version to 3.8, which is the oldest available in github-workflow-available linux versions).

I've also added a small check fixing the one flake8 issue.

With these changes, the workflow runs succesfully again, so I think this is ready to merge now.

PR #718, opened 2023-04-26
CI: Add link to flatpak build in PR comment

matthijskooijman:


matthijskooijman:
This PR adds a workflow that adds a comment to a pr with a link to the generated flatpak build whenever the pr is updated for easy testing. The workflow only works when it is in master, so it only works after this pr is merged.

To test it anyway, I pushed it to my own fork's master branch, you can see the comment generated here: https://github.com/matthijskooijman/hamster/pull/2#issuecomment-1524025333 (it took a few tries to get the formatting right, see that same pr for earlier versions).

There is still a "foo" testing commit in this PR that I will remove tomorrow, but othherwise I'm pretty happy with it.

Comments about this approach are welcome.

matthijskooijman:
I'm going to go ahead and merge this without further review, so we can start using it for more easily testing other PRs already. I'll still welcome review and opinions on how this works and am happy to further tweak how this works later.

PR #720, opened 2023-04-28
feat: add resume command

xai:
Support resuming the last tracked activity.

When `hamster resume` is called, the last tracked activity is started again with the current start_time.

If `hamster resume --continuous` (or short with `-c`) is called, the last tracked activity continues to be tracked, i.e., its end time is removed.


xai:
> In the case of --continuous, it might make more sense to simply update the last task removing the end time.

Great idea, I adjusted my commit accordingly.

DirkHoffmann:
Linguistically, I would rather go for `--continue` (or `--pick-up` in a more common way) than `--continuous`.

Some native speaker's view on this?

aquaherd:
Not a native speaker either but how about `hamster resume --no-gap`?

xai:
Sue, I can rebase and change the naming. I like `hamster resume --no-gap,` as this is exactly what it does. Would that be ok with you?

DirkHoffmann:
> Sue, I can rebase and change the naming. I like `hamster resume --no-gap,` as this is exactly what it does. Would that be ok with you?

I would still be in favour of `--continue`, even more because we decided to continue the previous entry by removing its end time rather than making a new one. For me, 
* `continue` evokes something like "11:00-12:00" → "11:00-now" and 
* `no gap` evokes rather "11:00-12:00" → "11:00-12:00 + 12:00-now"

for an ongoing task.

But we can also wait, until I asked a native speaker today.

xai:
Thanks, sounds good! Let me know when you have decided and I should rebase!

GeraldJansen:
I think separate actions, `hamster continue` and `hamster resume` would be more convenient and would be easier to document. 

The command line documentation, ie. `hamster help`, should be updated.

Also, the list of options in `hamster.bash` should be updated, i.e. `opts="activities categories current export list search start stop ".

matthijskooijman:
> I think separate actions, hamster continue and hamster resume would be more convenient and would be easier to document.

Sounds reasonable, though I would not be in favor of "continue" and "resume", since they really sound like they are the same thing.

Maybe "restart" is an option for the with-gap case?

DirkHoffmann:
After consultation, `resume --continue` would be best choice (semantically).

But I am actually more with @GeraldJansen's proposal for separate `hamster continue` and `hamster resume` commands instead of the optional parameter. I don't think that they are very different and the CLI is probably (?) not used directly by a majority of users. So it is rather a matter of documenting it correctly for those who write other interfaces.

The only argument to keep `hamster resume --continue` rather than `hamster continue` might be that they are pretty much too similar to appear as two distinct commands. (I don't know, if there are similar precedences for other commands.)

Anyway, I think the discussion went long enough. (Thanks for all the input!) And the person who actually implements the feature should be able now to make a sensible choice.

xai:
Thank you for taking the time to provide so much valuable feedback! I appreciate the varied perspectives on the command line interface change, which certainly has to be carefully reviewed.

After considering the discussion, it seems there are two primary viewpoints: one emphasizing the clarity of `resume` and `continue` as standalone commands (with appropriate documentation, which is now included), and another suggesting a more explicit option like `resume --no-gap` to address potential confusion due to similarities (like my original draft, but a lot clearer in what it does).

Considering these insights and our existing support for aliases in the command line interface, I've updated my pull request to accommodate both preferences. The revised implementation now includes separate commands for resume and continue, alongside adding a `--no-gap` option for the resume command.

I hope this adjustment aligns with the collective preferences. Please feel free to share any further thoughts or suggestions.

xai:
Thanks for all that valuable feedback! I understand the different viewpoints and would like to make everyone happy because changes to the command line interface should indeed be thought through well ;)

So I figure that `continue` and `resume` should be separate commands because they do different things, but that a `resume --no-gap` might still be interesting to start the most recent activity from its previous end time.

I will further revise my PR to incorporate all your feedback!

xai:
(Sorry, I misclicked and closed accidentally)

Now, we should have all three behaviors implemented and documented:

- `continue` removes the end time of the most recent activity
- `resume` starts the most recent activity with the current end time
- `resume --no-gap` starts the most recent activity with its previous end time

I'm looking forward to further suggestions!

xai:
Amended again to make flake8 happy as well.

aquaherd:
@xai Sorry to be a bother but could you please add your new commands to the bash completion?

xai:
> @xai Sorry to be a bother but could you please add your new commands to the bash completion?

No problem at all! I now also complete the `--no-gap` of the `resume` command (the commands themselves were already being completed).

After rebasing to the current master, I re-tested my commands (i.e., `continue`, `resume`, and `resume --no-gap`) and found that I had to add a fix for `hamster continue` to actually set the current activity by setting the end_time to `null` explicitly. Just calling `__touch_fact()` with a `None` end_time resulted in an updated end_time but not in active tracking. Therefore, I now added `__continue_fact()` in `db.py` and call that.

PR #727, opened 2023-09-04
Remove pyxdg dependency

a-andre:


matthijskooijman (approved):
It looks like [GLib.get_user_data_dir](https://docs.gtk.org/glib/func.get_user_data_dir.html) will also return XDG_DATA_HOME (which according to [XDG specs](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) defaults to ~/.local/share just like we did), so I think this PR will produce the same behavior as before, but with one less dependency (we already depend on glib), so looks good to me.

I'm approving this based on my code review, it should probably be tested by someone else (or me later) before actually merging.

matthijskooijman:
@aquaherd, I believe you've been using "rebase and merge" for merging some recent pull requests, right? I think we should prefer regular merging (or for single-commit PRs squash-merging), since then the commit history retains a reference to the pullrequest that a commit came from. See also https://github.com/projecthamster/hamster/issues/585 for some (draft - we should finalize that) development guidelines. No worries for now, just something to keep in mind for the future.

aquaherd:
Ok will do.

PR #733, opened 2023-10-01
Add Fedora dependencies

hedayat:


aquaherd:
Why are the instructions for 'rpm based' distributions no longer valid and are now fedora only?

IIRC, OpenSuSe still uses rpm.

hedayat:
> Why are the instructions for 'rpm based' distributions no longer valid and are now fedora only?
> 

No, actually they were never valid. And, you can already see that there is an OpenSUSE section right above this section, and it differs.

The actual problem is that RPM is not a distribution, it's just a packaging format. So, "RPM-based distros" have nothing to do with "Debian-based distros". AFAIK, there are no distros using DEB packaging format which is not derived from Debian, so the concept doesn't exist in DEB-based world (or, it is not well known).

RPM-based distros are using RPM, but that's it! They might have completely different package managers (DNF in Fedore vs Zypper in OpenSUSE), and even if they used the same package managers, the package sets can be completely different and use different naming conventions. So, a package name in Fedora can be XY, while in OpenSUSE it might by XYZ. 

Simply, Fedora and OpenSUSE have no parent/child relationship, and they have completely independent repositories.

aquaherd:
Ok thanks for clarifying. I will merge this if no other objections arise.

PR #736, opened 2023-10-16
Add missing screenshot of editor widget

GeraldJansen:
Page https://github.com/projecthamster/hamster/wiki has some screenshots but one is missing. This PR adds a screenshot of the editor window accessible from the wiki as a blob.


aquaherd:
Will merge this if no objections given.

matthijskooijman (approved):

PR #737, opened 2023-11-13
Fixed Improper Method Call: Replaced `NotImplementedError`

fazledyn-or:
## Details
While triaging your project, our bug fixing tool generated the following message(s)-

> In file: [datetime.py](https://github.com/projecthamster/hamster/blob/master/src/hamster/lib/datetime.py#L57), class: date, there is a special method `__sub__` that raises a NotImplementedError. If a special method supporting a binary operation is not implemented it should return NotImplemented. On the other hand, NotImplementedError should be raised from abstract methods inside user defined base classes to indicate that derived classes should override those methods. iCR suggested that the special method `__sub__` should return NotImplemented instead of raising an exception. An example of how NotImplemented helps the interpreter support a binary operation is here.

### Related Documentation
- [docs.python.org - NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented)
- [docs.python.org - NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError)
- [docs.python.org - Implementing Arithmetic Operators](https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations)

## Changes
- Replaced `NotImplementedError` with `NotImplemented`

## Previously Found & Fixed
- https://www.github.com/SciTools/iris/pull/5544
- https://www.github.com/cupy/cupy/pull/7900
- https://www.github.com/ethereum/web3.py/pull/3080

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/src/hamster/lib/datetime.py
+++ b/src/hamster/lib/datetime.py
@@ -54,7 +54,7 @@ class date(pdt.date):
         elif isinstance(other, pdt.date):
             return timedelta.from_pdt(self.to_pdt() - other)
         else:
-            raise NotImplementedError("subtract {}".format(type(other)))
+            return NotImplemented
 
     @classmethod
     def parse(cls, s):
@@ -250,7 +250,7 @@ class datetime(pdt.datetime):
         elif isinstance(other, datetime):
             return timedelta.from_pdt(self.to_pdt() - other)
         else:
-            raise NotImplementedError("subtract {}".format(type(other)))
+            return NotImplemented
 
     def __str__(self):
         if self.tzinfo:


fazledyn-or:
> Are we accepting AI-generated patches for review? Is this spam?

No. This is manually done by me. 
However, seems like I made a mistake. I should've used `TypeError` instead of `NotImplemented`. Let me fix it ASAP.

GeraldJansen:
Sure smells like spam to me. @fazledyn-or, you have made dozens of very similar, trivial pull requests in just as many projects over the last month or so. What gives?

fazledyn-or:
We are working with the OpenSSF under the Linux Foundation to scan top 10,000 Java, Python and Go projects and find and fix bugs in them. As a part of that effort, in the recent past, I have engaged with the maintainers of many Python projects. The bugs reported include XSS, crypto issues, deserialization, improper method call, etc.

At the same time, we have also reported some simple and easy to fix issues such as this bug. In this case, there was a mistake in the fix we suggested at first. I have corrected it. It is up to the maintainers whether to accept the simple (but appropriate) fix that has been proposed here.

We use our SAST tool to scan the projects, then based on the results, we clone the project locally and triage ourselves. And then manually create the commits and submit PR. None of it is done by bots.

matthijskooijman:
@fazledyn-or Thanks for clarifying and submitting this change!

I've reviewed the docs about this, and it seems indeed that returning `NotImplemented` instead of raising `NotImplementedError` is required, according to [this documentation](https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations). I wondered if this maybe changed between py2 and py3 (and thus the current hamster code would be legacy from py2 code), but it seems [py2 has the same documentation, requiring returning `NotImplemented`](https://docs.python.org/2/library/numbers.html#implementing-the-arithmetic-operations). So it seems like a reasonable change to accept, based on the documentation links provided.

However, @fazledyn-or in your recently added commit, you change from returning `NotImplemented` to returning `TypeError`, which I cannot see documented anywhere (and it seems `TypeError` is intended to be raised, not returned). So why did you make this change?

fazledyn-or:
According to the Python [documentation](), `NotImplemented` should be returned. However, my last change on using `TypeError` was due to the fact that `__sub__` operation in `date` is only supported with a few classes. Even the TypeError [documentation](https://docs.python.org/3/library/exceptions.html#TypeError) says that too.

However, using `return NotImplemented` results in a `TypeError` too (after the Python interpreter tries to perform the __sub__ operation and fails). I used `TypeError` here so that the exception can be seen explicitly in the source code, than during run-time.

And yes, I made another mistake (using return instead of raising the exception). For that I'm embarrassed and sincerely apologize.

matthijskooijman:
> And yes, I made another mistake (using return instead of raising the exception). For that I'm embarrassed and sincerely apologize.

Right,  raising it would make more sense indeed.

> However, using return NotImplemented results in a TypeError too (after the Python interpreter tries to perform the sub operation and fails). I used TypeError here so that the exception can be seen explicitly in the source code, than during run-time.

I have also read (but only in passing) that the design is to return NotImplemented rather than raising TypeError directly, I think so the code that calls `__add__` can potentially try the reverse operation as well (in case of adding mixed types), and presumably because raising an exception prevents this. This is not applicable for `__sub__` (since it cannot be reversed), but maybe there is still a reason the documentation says to return `NotImplemented` rather than raising.

So I'm inclined to prefer the original version, which seems to conform to documentation better (also, it's very much a corner case that I do not think we really need to care very much about at all...). Would that make sense?

fazledyn-or:
That would make sense. I have reverted the commit back to `return NotImplemented`. Thank you for your cooperation and kind explanation. It really helps us understand the impact of these bug fixes.

matthijskooijman (approved):
With the recent change reverted, looks good to merge like this for me.

PR #739, opened 2023-11-19
flatpak: Update gnome platform

matthijskooijman:
The previously used version 41 has been unsupported since 2022, so update it.

This does not update dbus-python yet, since the 1.3.0 version requires meson as a build system, but seems to require a newer version than what the gnome/freedesktop 45 SDK supports and manually installing a newer version takes more effort with no signficant gains, see also https://github.com/flathub/net.lutris.Lutris/blob/7bd51222b8076abd8fcbfe9cb0e4d6e70bafca24/net.lutris.Lutris.yml#L775-L780


matthijskooijman:
Since this is rather trivial, I'm going to just proceed and merge this right way.

PR #749, opened 2023-12-03
A few small fixes

### LINKED ISSUE #747: Small fixes
Just a few small fixes that I noticed during the 3.0.3 release, that could be made or investigated:
 - [ ] Move dbus service files into `data/` or another subdir for cleanup. Also maybe move the flatpak file, see #743.
 - [ ] Drop LIBDIR from defs.py.in (for Debian reproducible builds), and maybe others too (only VERSION and DATADIR seems used).
 - [x] ~~Is `data/hamster.db` actually used? Does it define the initial db schema, or is it leftover from an old version?~~ Yes, still used.
 - [ ] See if we can remove `distutils` import (it seems deprecated, and according to lintian removed in Python 3.12, though it does still seem to work in 3.12 so far).
 - [ ] Strip whitespace from VERSION to allow newlines (https://github.com/projecthamster/hamster/pull/742#issuecomment-1817954566)
 - [ ] Fix py3.6 compatibility for `subprocess.run`, see https://github.com/projecthamster/hamster/pull/738#pullrequestreview-1740265113

matthijskooijman:
This PR collects a number of small fixes for problems noticed around the previous release, as collected in #747, plus a few other things.

In particular, it applies some fixes to the version tracking code introduced for the 3.0.3 release, removes some unused variables to make the build more reproducible and moves some files out of the repository root.

Taken together, this PR fixes #747.


rhertzog:
FTR, if you wanted to have Python version parsing/comparison, I believe the current/modern way is to use the "packaging" module (apt install python3-packaging on Debian).

I did review your changes and they look good to me. I believe this can be merged.

matthijskooijman:
> FTR, if you wanted to have Python version parsing/comparison, I believe the current/modern way is to use the "packaging" module (apt install python3-packaging on Debian).

Yeah, I saw that, but it did seem worth adding an external dependency just for this check.

> I did review your changes and they look good to me. I believe this can be merged.

Thanks!

PR #760, opened 2024-01-29
Remove dependency on deprecated distutils package.

GeraldJansen:
As of Python 3.10 the distutils package has been deprecated and it will be removed from Python 3.12. This PR uses alternative functions from the packaging.versions submodule.


matthijskooijman:
Seems you're working with an out-of-date version of the master branch locally, this has already been resolved in #749.

GeraldJansen:
Oops, already fixed.

PR #763, opened 2024-07-04
Fix today date change

### LINKED ISSUE #633: Today always links to day of app start
at least in the snap package.

MarSoft:
This fixes #633 
Things to improve: when the user changes day start in config, we should probably perform `update_today` too. Not sure though; anyway, this situation is not so frequent.


MarSoft:
Using the version with this PR applied every day for more than 3 months, works fine.

